PR #1, opened 2020-02-12
Fixes MAISTRA-1168: added implementation of boringssl function SSL_get_peer_full_cert_chain

dmitri-d:
This fixes MAISTRA-1168.

The issue was that the openssl call (SSL_get_peer_cert_chain) that we used to substitute a boringssl one (SSL_get_peer_full_cert_chain) doesn't return a full certificate chain when called on the server: client peer certificate needs to be retrieved separately (see https://www.openssl.org/docs/man1.1.0/man3/SSL_get_peer_cert_chain.html for more details). I added an implementation of SSL_get_peer_full_cert_chain that emulates boringssl behaviour.

### DIFF
--- a/source/extensions/transport_sockets/tls/openssl_impl.cc
+++ b/source/extensions/transport_sockets/tls/openssl_impl.cc
@@ -90,6 +90,24 @@ std::string getSerialNumberFromCertificate(X509* cert) {
   return "";
 }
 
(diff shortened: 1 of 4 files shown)


dmitri-d:
@dgn: Yep, noticed it too.

PR #3, opened 2020-02-16
Fixes MAISTRA-1167: RetryHostPredicateFilter in protocol_integration_test suite passes now

dmitri-d:

### DIFF
--- a/test/test_common/utility.h
+++ b/test/test_common/utility.h
@@ -444,7 +444,7 @@ public:
   static std::string convertTime(const std::string& input, const std::string& input_format,
                                  const std::string& output_format);
 
-  static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(30000);
+  static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(10000);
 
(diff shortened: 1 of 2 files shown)


test/test_common/utility.h
```diff
@@ -444,7 +444,7 @@ class TestUtility {
   static std::string convertTime(const std::string& input, const std::string& input_format,
                                  const std::string& output_format);
 
-  static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(30000);
+  static constexpr std::chrono::milliseconds DefaultTimeout = std::chrono::milliseconds(10000);
```
brian-avery: What was the failure in this test and why does decreasing the timeout fix it?

dmitri-d: It was failing with an error that a connection has been torn down unexpectedly. The issue was with test being torn down before the test code had a chance to handle connection timeout. For some reason ```DefaultTimeout``` was increased in istio-envoy repository (I don't think this value will work at all). This resets ```DefaultTimeout``` back to its original value.

dmitri-d:
istio-proxy tests are passing with this change.

PR #7, opened 2020-03-12
Fixes MAISTRA-1265: Upstream release 1.4.6 + alpn fixes + latest bssl_wrapper

dmitri-d:
This PR 
* brings maistra-1.1 branch up to date with changes in upstream 1.4.6 
* updates bssl_wrapper depedency to latest
* resolves alpn issues in tls_inspector

I decided to postpone moving code accessing SSL struct internal into bssl_wrapper until the next release, mostly to simplify and speed up 1.1 release.

### DIFF
--- a/bazel/external/openssl_includes-1.patch
+++ b/bazel/external/openssl_includes-1.patch
@@ -1,0 +1,13 @@
+diff --git a/ssl/packet_locl.h b/ssl/packet_locl.h
+index 860360b8b2..49c719285f 100644
+--- a/ssl/packet_locl.h
++++ b/ssl/packet_locl.h
+@@ -426,7 +426,7 @@ __owur static ossl_inline int PACKET_memdup(const PACKET *pkt,
+     if (length == 0)
+         return 1;
+ 
+-    *data = OPENSSL_memdup(pkt->curr, length);
++    *data = (unsigned char *)OPENSSL_memdup(pkt->curr, length);
+     if (*data == NULL)
+         return 0;
+ 
--- a/bazel/external/openssl_includes.BUILD
+++ b/bazel/external/openssl_includes.BUILD
@@ -1,0 +1,18 @@
+cc_library(
+    name = "openssl_includes_lib",
+    copts = ["-Wno-error=error"],
+    hdrs = [
+            "e_os.h", 
+            "ssl/ssl_locl.h", 
+            "ssl/packet_locl.h", 
+            "ssl/record/record.h", 
+            "ssl/statem/statem.h", 
+            "include/internal/dane.h",
+            "include/internal/nelem.h",
+            "include/internal/numbers.h",
+            "include/internal/refcount.h",
+            "include/internal/tsan_assist.h",
+    ],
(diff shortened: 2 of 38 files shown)


bazel/external/openssl_includes-1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/ssl/packet_locl.h b/ssl/packet_locl.h
+index 860360b8b2..49c719285f 100644
+--- a/ssl/packet_locl.h
++++ b/ssl/packet_locl.h
+@@ -426,7 +426,7 @@ __owur static ossl_inline int PACKET_memdup(const PACKET *pkt,
+     if (length == 0)
+         return 1;
+ 
+-    *data = OPENSSL_memdup(pkt->curr, length);
++    *data = (unsigned char *)OPENSSL_memdup(pkt->curr, length);
```
dmitri-d: This is to make g++ compiler happy (implicit casts from void* are illegal in c++): I couldn't find a way to make bazel use gcc for this dependency.

bazel/external/openssl_includes.BUILD
```diff
@@ -0,0 +1,18 @@
+cc_library(
```
dmitri-d: This is used to open SSL structure to allow access to alpn-related fields here: https://github.com/Maistra/envoy/pull/7/files#diff-399c9deb5d9027103b258cd84614c9b6R59

We continue to link against system ssl libraries, this is just to allow inclusion of internal openssl headers.

bazel/repository_locations.bzl
```diff
@@ -23,9 +23,9 @@ REPOSITORY_LOCATIONS = dict(
     ),
     #EXTERNAL OPENSSL
     bssl_wrapper = dict(
-        sha256 = "81a59d013096015a93269325ee4148d826ffd7a9f019f622850a2b86974b9748",
-        strip_prefix = "bssl_wrapper-2eaed8832e12a0fada8f08a5e23522e035b80784",
-        urls = ["https://github.com/maistra/bssl_wrapper/archive/2eaed8832e12a0fada8f08a5e23522e035b80784.tar.gz"],
+        sha256 = "d84ea7d190210145695e5b172e8e6fb23f3464360da5efab5a1ae1a973c21f57",
+        strip_prefix = "bssl_wrapper-c9649facde3ab1d8bc871c7375a8946c50950e97",
+        urls = ["https://github.com/maistra/bssl_wrapper/archive/c9649facde3ab1d8bc871c7375a8946c50950e97.tar.gz"],
```
brian-avery: Out of curiosity, why don't you use the commit as a variable here?

dmitri-d: I'm not sure it's possible in .bzl?

jwendell: They are just python files. You can assign a variable and refer it later. That said, I think we should follow the standard on the file. If the rest of file is doing this way, let's just do the same.

brian-avery: True, handling this as a variable would be in line with the way proxy does it. Not sure what the standard is here.

PR #27, opened 2020-06-18
Fixes MAISTRA-1564: errno is now being logged when SSL_read() fails

dmitri-d:


knrc:
> LGTM
> Is this used only for debug purposes? I mean, should we worry about any performance impact? I'm assuming `ENVOY_CONN_LOG` is a macro that becomes a noop when debug is disabled...

It's not a noop but is guarded by the level check and would only appear for SSL_ERROR_SYSCALL failures.

dmitri-d:
>  I'm assuming ENVOY_CONN_LOG is a macro that becomes a noop when debug is disabled...
This is how the macro is defined: https://github.com/Maistra/envoy/blob/maistra-1.2/source/common/common/logger.h#L302

The performance impact is minimal, vast majority of cases will skip the log altogether.

PR #28, opened 2020-06-18
Fixes MAISTRA-1564: errno is now being logged when SSL_read() fails

dmitri-d:
backport of https://github.com/Maistra/envoy/pull/27


dmitri-d:
@knrc: ping re: block.

PR #38, opened 2020-08-20
Add script to be used by CI

jwendell:


maistra/run-ci.sh
```diff
@@ -0,0 +1,24 @@
...
+export BUILD_SCM_STATUS="SHA=${PULL_PULL_SHA:-undefined}"
+
+# Build
+bazel build //source/exe:envoy-static
+
+echo "Build succeeded. Binary generated:"
+bazel-bin/source/exe/envoy-static --version
+
+# Run tests
+bazel test //test/...
```
dmitri-d, line 24: Could you add --build_tests_only here? Otherwise this will attempt to build everything under /test (e.g. fuzzer tests, which are not based on gunit): which is slower at best, may fail at worst.

jwendell: Done

PR #40, opened 2020-09-09
MAISTRA-1810: check timeout period to ensure the delay is relative to…

knrc:


dmitri-d:
I don't understand why we are seeing such a clock drift. I don't think this issue is some sort of a race -- all calls are made from within the same thread. 

Handling of timers in libevent is quite involved (I tried asking the question on their irc but no response so far), but I wonder if we are seeing an interaction between cached time, their use of `gettimeofday` and `clock_gettime` calls, and the way the drift between the last two is calculated/used. We could enable debug logging of event  handling (would show handling of timeout events) and/or reduce the interval which is used for syncing of `gettimeofday` and `clock_gettime` from 5s to maybe 1s?

I'll post the question on their mailing list. Is the cluster running under a hypervisor?

knrc:
@dmitri-d I don't believe clock drift is the issue, nor threading, but rather the sequence of events with the timer and dispatch.  My theory is that the first write sets the timer, this then expires and queues up the handler, in the meantime another write event takes place (should reset the timer) followed shortly after by the timer handler.  If we are sticking to the documentation then the handler should have been cancelled and restarted.

dmitri-d:
>  If we are sticking to the documentation then the handler should have been cancelled and restarted.

Yep, and I can see it in the libevent code. I only just realised that we don't check the return value of the `event_add` call, which can be an error (-1); in case of an error the internal state of the event loop isn't changed, which could explain missed update.

dmitri-d (approved):

dmitri-d:
There's one test failure, but I don't think it's related to this change, but rather is a falky test.

PR #47, opened 2020-10-08
MAISTRA-1921: Support dynamic crypto based on environment

knrc:
This PR modifies the defaults used when running within a FIPS environment.

### DIFF
--- a/source/extensions/transport_sockets/tls/context_config_impl.cc
+++ b/source/extensions/transport_sockets/tls/context_config_impl.cc
@@ -358,7 +358,7 @@ ClientContextConfigImpl::ClientContextConfigImpl(
 const unsigned ServerContextConfigImpl::DEFAULT_MIN_VERSION = TLS1_VERSION;
 
 // FIPS configuration
-const unsigned ServerContextConfigImpl::DEFAULT_FIPS_MAX_VERSION = TLS1_3_VERSION;
+const unsigned ServerContextConfigImpl::DEFAULT_FIPS_MAX_VERSION = TLS1_2_VERSION;
 const std::string ServerContextConfigImpl::DEFAULT_FIPS_CIPHER_SUITES =
     "ECDHE-ECDSA-AES128-GCM-SHA256:"
     "ECDHE-RSA-AES128-GCM-SHA256:"
@@ -375,7 +375,7 @@ const std::string ServerContextConfigImpl::DEFAULT_FIPS_CIPHER_SUITES =
 const std::string ServerContextConfigImpl::DEFAULT_FIPS_CURVES =
     "P-256";
 // Non FIPS configuration
-const unsigned ServerContextConfigImpl::DEFAULT_NON_FIPS_MAX_VERSION = TLS1_2_VERSION;
+const unsigned ServerContextConfigImpl::DEFAULT_NON_FIPS_MAX_VERSION = TLS1_3_VERSION;
 const std::string ServerContextConfigImpl::DEFAULT_NON_FIPS_CIPHER_SUITES =
     "[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]:"
     "[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]:"
--- a/source/extensions/transport_sockets/tls/context_config_impl.cc
+++ b/source/extensions/transport_sockets/tls/context_config_impl.cc
@@ -1,5 +1,6 @@
 #include "extensions/transport_sockets/tls/context_config_impl.h"
 
+#include <fstream>
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


jwendell (approved):

brian-avery (approved):

source/extensions/transport_sockets/tls/context_config_impl.cc
```diff
@@ -19,6 +21,30 @@ namespace Tls {
 
 namespace {
 
+absl::once_flag fipsOnce;
+bool fipsEnabled;
+
+void initFipsEnabled() {
+  std::ifstream file("/proc/sys/crypto/fips_enabled");
+  if (file.fail()) {
+    std::cout << "FIPS " << (fipsEnabled ? "enabled" : "disabled") << "\n";
```
jwendell, line 30: FIPS is always disabled here. What's the content of `fipsEnabled` variable at this point? Better to go back to a static message?

source/extensions/transport_sockets/tls/context_config_impl.cc
```diff
@@ -19,6 +21,30 @@ namespace Tls {
...
+
+    std::string fipsEnabledText = file_string.str();
+    fipsEnabledText.erase(fipsEnabledText.find_last_not_of("\n") + 1);
+    fipsEnabled = fipsEnabledText.compare("1") == 0;
+    std::cout << "FIPS " << (fipsEnabled ? "enabled" : "disabled") << ": " << fipsEnabledText << "\n";
+  }
+}
+
+bool isFipsEnabled() {
+  absl::call_once(fipsOnce, initFipsEnabled);
```
jwendell, line 44: Tricky :) In the old good days we would just use a `static` variable that would store whether or not we need to run something :)

source/extensions/transport_sockets/tls/context_config_impl.cc
knrc: True, this was a static message and should be reverted to that

source/extensions/transport_sockets/tls/context_config_impl.cc
knrc: It seemed safer than playing directly with mutexes :)

knrc: and statics don't work in certain memory models and regularly break

source/extensions/transport_sockets/tls/context_config_impl.cc
```diff
@@ -282,7 +308,24 @@ unsigned ContextConfigImpl::tlsVersionFromProto(
 const unsigned ClientContextConfigImpl::DEFAULT_MIN_VERSION = TLS1_VERSION;
 const unsigned ClientContextConfigImpl::DEFAULT_MAX_VERSION = TLS1_2_VERSION;
 
-const std::string ClientContextConfigImpl::DEFAULT_CIPHER_SUITES =
+// FIPS configuration
+const std::string ClientContextConfigImpl::DEFAULT_FIPS_CIPHER_SUITES =
```
dmitri-d, line 312: Both server and client default cipher suites look the same. I think they are going to stay this way too. Any reason not to move these to the base class then?

source/extensions/transport_sockets/tls/context_config_impl.cc
```diff
@@ -19,6 +21,30 @@ namespace Tls {
 
 namespace {
 
+absl::once_flag fipsOnce;
+bool fipsEnabled;
+
+void initFipsEnabled() {
+  std::ifstream file("/proc/sys/crypto/fips_enabled");
+  if (file.fail()) {
+    std::cout << "FIPS disabled\n";
```
dmitri-d, line 30: Could you replace cout's with `ENVOY_LOG(debug, "message here")`? You'll need to include `"common/common/logger.h"` and add `"//source/common/common:logger_lib"` dependency to the `context_lib` target in `source/extensions/transport_sockets/tls/BUILD`.

knrc: No, I had that initially and it's at warning level so we don't see it.

knrc: I don't want to set it at warning, or higher, to guarantee it is displayed every time

source/extensions/transport_sockets/tls/context_config_impl.cc
knrc: With the exception of the TLS max they are but this just follows the upstream behaviour which I'm not changing

knrc:
I'm going to revert the code tomorrow, back to the logging version, and then try to bypass the level settings to always display

source/extensions/transport_sockets/tls/context_config_impl.cc
dmitri-d: Would info level work here?

knrc:
I had a long think about this while on a bike ride

The following options have been tried
- variable initialiser with logging framework (impacts some fragile Envoy tests which break with the extra output since logging levels have not been set to warning)
- on demand initialisation with logging framework (logging level is at warning so no output)
- on demand initialisation with cout - Success but not ideal solution

since each of these causes issues, or is undesirable in some way, I've decided to revert to the original variable initialisation and not log the status.

knrc:
Test failure is IpVersions/IntegrationTest.ResponseFramedByConnectionCloseWithReadLimits/IPv4 which seems to be a flaky test, it has happened with previous PRs.

jwendell:
> I had a long think about this while on a bike ride
> 
> The following options have been tried
> 
>     * variable initialiser with logging framework (impacts some fragile Envoy tests which break with the extra output since logging levels have not been set to warning)
> 
>     * on demand initialisation with logging framework (logging level is at warning so no output)
> 
>     * on demand initialisation with cout - Success but not ideal solution
> 
> 
> since each of these causes issues, or is undesirable in some way, I've decided to revert to the original variable initialisation and not log the status.

If you wish, you can append something like "/FIPS" in the output of `--version` command here: https://github.com/maistra/envoy/blob/maistra-1.1/source/common/common/version.cc#L22

knrc:
@jwendell that's a good idea, although at this point I've spent more time on fixing the logging than I did on fixing the actual issue

source/extensions/transport_sockets/tls/context_config_impl.cc
knrc: No, that's what I had but the default settings are for warning and above.  @jwendell came up with a better solution and I'll try to implement that in a separate PR.

PR #50, opened 2020-10-13
MAISTRA-1934: Support dynamic crypto based on environment

knrc:


jwendell (approved):

jwendell:
timeout of 5h :o

dmitri-d:
Looks like the job failure is due to a prow timeout.

jwendell:
This job used to run in ~4h. Now 5h is not enough. I wonder what happened.. our prow k8s cluster just got slower?

Anyway, I'll increase it to 6h. That's just insane :(

PR #68, opened 2021-02-05
Patch wasm sdk and use host's nodejs

jwendell:


dmitri-d (approved):

oschaaf (approved):
Looks good too me, but I have two questions:
- I wonder, would it be worth it to add the rationale for this change to the commit description? 
- Also, right now this repo needs a single modification (change [update line](https://github.com/maistra/envoy/blob/b6af8fd72ae66e387da27aca1f45d2f3a65b28fc/WORKSPACE#L25)) to get it to build on my (ubuntu) dev box. I suspect that upon merging this more changes will be needed, is that right? Would it be worth it to track modifications that constrain the capacity to build on different platforms somewhere?

jwendell:
@oschaaf This is part of the work to build WASM in our build system. Dmitri and I usually discuss this over IRC, I'll try to be more verbose on the commit messages.

Also, Envoy/Proxy are meant to be built using our [build image](https://github.com/maistra/test-infra/blob/main/docker/maistra-proxy-builder_2.1.Dockerfile). They also might compile on our host, but we use Fedora, not Ubuntu :)

PR #72, opened 2021-02-11
Remove LLVM dependency

jwendell:
We don't use it. It saves ~500MB on disk / ~50MB on tarball.


bazel/foreign_cc/BUILD
```diff
@@ -197,109 +197,6 @@ envoy_cmake_external(
     }),
 )
 
-envoy_cmake_external(
-    name = "llvm",
```
dmitri-d, line 201: There's a dependency on the `llvm` target here: https://github.com/maistra/envoy/blob/44abadf7a67d8f0c83676108ce8c84471b635cf7/bazel/foreign_cc/BUILD#L245

jwendell: @dmitri-d yeah, I saw that. Apparently it's not used too. I was able to build (locally) both envoy and proxy without llvm.

dmitri-d: how weird!

dmitri-d (approved):

PR #86, opened 2021-04-09
Fixes MAISTRA-2119: certificate stapling is now functional under OpenSSL

dmitri-d:
- updated bssl_wrapper dependency to the latest (required for this
  change)
- removed cbs dependecy (now part of bssl_wrapper)
- test/extensions/transport_sockets/tls/ocsp/test_data/ directory has
  been synced with upstream, as some/all certs there have expired
- ditto test/extensions/transport_sockets/tls/test_data/
- When multiple certs chains are configured for a TLS context,
  it's currently possible to end up with SSL handshake failing if
  a client requests an OCSP response, but OpenSSL selects a cert chain
  with an invalid/missing OCSP response.
  See https://issues.redhat.com/browse/MAISTRA-2258.

Note this PR requires https://github.com/maistra/jwt_verify_lib/pull/1 to be merged, build will fail until then.


source/extensions/transport_sockets/tls/context_impl.h
```diff
@@ -264,9 +256,11 @@ class ServerContextImpl : public ContextImpl, public Envoy::Ssl::ServerContext {
                          unsigned int inlen);
   int sessionTicketProcess(SSL* ssl, uint8_t* key_name, uint8_t* iv, EVP_CIPHER_CTX* ctx,
                            HMAC_CTX* hmac_ctx, int encrypt);
-  OcspStapleAction ocspStapleAction(const ServerContextImpl::TlsContext& ctx,
+  bool isClientOcspCapable(SSL* ssl);
+  const ContextImpl::CertContext& certificateContext(X509* cert);
```
oschaaf, line 260: These are specific to our implementation right? Should we consider documenting any new method(s)? This would then also be a good  place to call out the assumptions around inputs (eg. when this asserts). Tangentially related, I'm curious, we `RELEASE_ASSERT` for certain inputs of `cert` here in the implementation of this, would this be a candidate for returning an `absl::status_or` instead and handling it in `ServerContextImpl::handleOcspStapling`? If this is truly invariant, then probably not, but in that case I'm wondering if there are tests in place around verifying that.

test/extensions/transport_sockets/tls/ocsp/test_data/certs.sh
```diff
@@ -4,23 +4,21 @@
...
+
+cd "$HERE" || exit 1
 trap cleanup EXIT
+
 cleanup() {
-  rm -f ./*_index*
-  rm -f ./*.csr
-  rm -f ./*.cnf
-  rm -f ./*_serial*
+    rm -f ./*.cnf
```
oschaaf, line 14: nit: indentation

test/extensions/transport_sockets/tls/BUILD
```diff
@@ -149,28 +140,44 @@ envoy_cc_test_library(
...
+#        "//include/envoy/server:transport_socket_config_interface",
+#        "//include/envoy/ssl/private_key:private_key_config_interface",
+#        "//include/envoy/ssl/private_key:private_key_interface",
+#        "//source/common/config:utility_lib",
+#        "//source/common/protobuf:utility_lib",
+#        "@envoy_api//envoy/extensions/transport_sockets/tls/v3:pkg_cc_proto",
+#    ],
+#)
+
+#envoy_cc_test(
```
oschaaf, line 164: What's happening with these tests? Should we comment about why we don't enable these tests?

oschaaf (commented):
Some whitespace nits, and two (probably naive) questions.

source/extensions/transport_sockets/tls/context_impl.h
dmitri-d: > These are specific to our implementation right?

They are. Although upstream has `isClientOcspCapable(), here: https://github.com/envoyproxy/envoy/blob/release/v1.16/source/extensions/transport_sockets/tls/context_impl.cc#L1391` which is similar in intent, but relies on BoringSSL api. In this case it's an API that doesn't have an OpenSSL analogue (parsing of SSL handshake), so I didn't bother trying to extend bssl wrapper with an implementation for it.

> Should we consider documenting any new method(s)?

These two are trivial, and I tried to use method names that are self-explanatory. I added comments to new/modified data-structures (`certificateContext()` retrieves a reference to a Maistra-specific one, for example) however.

> Tangentially related, I'm curious, we RELEASE_ASSERT for certain inputs of cert here

Not entirely sure if you are talking about this: https://github.com/maistra/envoy/pull/86/files#diff-2b39f212c2f760ae547351a46f92ad08c0ec84c556d678ebe9388b9d008e7313R1290. If so: the assert here is b/c we should never ever *not* match a certificate: it's selected by OpenSSL from certs configured for a given TLS context (here's the lookup map being populated: https://github.com/maistra/envoy/pull/86/files#diff-2b39f212c2f760ae547351a46f92ad08c0ec84c556d678ebe9388b9d008e7313R313, a bit earlier there the cert chain itself is being loaded into the SSL_CTX). Not finding a match wouldn't be a runtime error, but a symptom of something being seriously wrong with the implementation.

> If this is truly invariant, then probably not, but in that case I'm wondering if there are tests in place around verifying that.

No additional tests, but I don't think we generally add tests to verify asserts (again, this isn't to catch a runtime issue).

test/extensions/transport_sockets/tls/BUILD
dmitri-d: There are tickets open for these: https://issues.redhat.com/browse/MAISTRA-2120 and https://issues.redhat.com/projects/MAISTRA/issues/MAISTRA-2121. I know this is ugly, but it makes it easier to merge in upstream changes (if marginally).

test/extensions/transport_sockets/tls/ocsp/test_data/certs.sh
dmitri-d: That's coming from upstream, I'd rather keep it (to avoid merge conflicts in the future).

source/extensions/transport_sockets/tls/context_impl.h
oschaaf: Thank you for explaining, that is very helpful. Based on that I can see `RELEASE_ASSERT` is the right choice here, and covering it in a test doesn't make sense (as it is invariant). 
As for doc comments, personally I tend try to err on the "generous" side (assuming naivity of future readers) but that is subjective, so up to you.

test/extensions/transport_sockets/tls/ocsp/test_data/certs.sh
oschaaf: makes sense. also, feel free to ignore these formatting nits, let me know if I should just ignore formatting in reviews (it's habitual for me to comment about it, but I can see that it might not always add value with what we're doing in this repo).

test/extensions/transport_sockets/tls/BUILD
oschaaf: fair enough

oschaaf (approved):
LGTM

dmitri-d:
How interesting, stats_integration_test, `MemoryLargeClusterSize` failed. 

/retest

test/extensions/transport_sockets/tls/ocsp/test_data/certs.sh
dmitri-d: Nah, thanks for catching these; Most of them were tabs injected by vim, good to get rid of those.

PR #87, opened 2021-04-16
1.14 changes

dmitri-d:
Cherry-pick of CVE fixes from istio-envoy/release-1.6 and an additional fix for handling of unknown TLS alert codes.


test/common/grpc/common_test.cc
```diff
@@ -99,8 +102,18 @@ TEST(GrpcContextTest, GetGrpcTimeout) {
   Http::TestRequestHeaderMapImpl unit_nanoseconds{{"grpc-timeout", "12345678n"}};
   EXPECT_EQ(std::chrono::milliseconds(13), Common::getGrpcTimeout(unit_nanoseconds));
 
-  // Max 8 digits and no leading whitespace or +- signs are not enforced on decode,
-  // so we don't test for them.
+  Http::TestRequestHeaderMapImpl value_overflow{{"grpc-timeout", "6666666666666H"}};
+  EXPECT_EQ(std::chrono::milliseconds(0), Common::getGrpcTimeout(value_overflow));
+
+  // Reject negative values.
+  Http::TestRequestHeaderMapImpl value_negative{{"grpc-timeout", "-1S"}};
```
oschaaf, line 109: Ah, yeah, so here starting with a `-` is covered in test, but what happens with superlarge negative valuesis not (and it might be worth doing that for all the units involved considering the trickery with std::durations?).

test/common/http/http2/http2_frame.h
```diff
@@ -125,6 +136,9 @@ class Http2Frame {
   // header.
   void appendHpackInt(uint64_t value, unsigned char prefix_mask);
   void appendData(absl::string_view data) { data_.insert(data_.end(), data.begin(), data.end()); }
+  void appendDataAfterHeaders(std::vector<uint8_t> data) {
+    std::copy(data.begin(), data.end(), data_.begin() + 9);
```
oschaaf, line 140: The +9 might benefit form a comment

source/common/grpc/common.cc
```diff
@@ -160,12 +160,18 @@ std::chrono::milliseconds Common::getGrpcTimeout(const Http::RequestHeaderMap& r
...
   const Http::HeaderEntry* header_grpc_timeout_entry = request_headers.GrpcTimeout();
   if (header_grpc_timeout_entry) {
-    uint64_t grpc_timeout;
-    // TODO(dnoe): Migrate to pure string_view (#6580)
-    std::string grpc_timeout_string(header_grpc_timeout_entry->value().getStringView());
-    const char* unit = StringUtil::strtoull(grpc_timeout_string.c_str(), grpc_timeout);
-    if (unit != nullptr && *unit != '\0') {
-      switch (*unit) {
+    int64_t grpc_timeout;
+    absl::string_view timeout_entry = header_grpc_timeout_entry->value().getStringView();
```
oschaaf, line 164: random thought: `absl::SimpleAtoi` below allows/ignores whitespace, did someone check the definition of whitespace matches the one that the trimming in the header parsing takes care of? It might be defensive to trim ourselves here, otherwise some of the assumptions about first/last characters might leave wiggle room for nasty edge cases.

oschaaf (commented):
A few questions, assuming those are non-issues, LGTM

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -263,9 +263,10 @@ ContextImpl::ContextImpl(Stats::Scope& scope, const Envoy::Ssl::ContextConfig& c
     if (tls_context_.cert_chain_ == nullptr ||
         !SSL_CTX_use_certificate(tls_context_.ssl_ctx_.get(), tls_context_.cert_chain_.get())) {
       while (uint64_t err = ERR_get_error()) {
-        ENVOY_LOG_MISC(error, "SSL error: {}:{}:{}:{}", err, ERR_lib_error_string(err),
-                       ERR_func_error_string(err), ERR_GET_REASON(err),
-                       ERR_reason_error_string(err));
+	ENVOY_LOG_MISC(debug, "SSL error: {}:{}:{}:{}", err,
```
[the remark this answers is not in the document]
dmitri-d: This is taken verbatim from upstream: https://github.com/envoyproxy/envoy/commit/ada478458a989d67ee452dc79ba279a99c01f23a#diff-2b39f212c2f760ae547351a46f92ad08c0ec84c556d678ebe9388b9d008e7313L299. My thinking is that b/c there's an exception raised immediately after, this log message isn't that useful under normal circumstances.

source/common/grpc/common.h
```diff
@@ -169,6 +169,8 @@ class Common {
 
 private:
   static void checkForHeaderOnlyError(Http::ResponseMessage& http_response);
+
+  static constexpr size_t MAX_GRPC_TIMEOUT_VALUE = 99999999;
```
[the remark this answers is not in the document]
dmitri-d: That's a max *value*. The timeout string is postfixed with a letter denoting the untis (i.e "H", "M", .., "u'. See https://github.com/maistra/envoy/blob/760ec40eaebd88bbfe3fb13945cb7ef1c8b4726a/source/common/grpc/common.cc#L175. A high value makes sense for micro- and nano-second intervals, I think).

dmitri-d:
Thanks for the review; In general I prefer not to make any changes in upstream code (including comments), unless we need modification to make it work for Maistra, The reasoning behind is that the more we change upstream code, the harder it becomes to merge and maintain.

oschaaf:
> Thanks for the review; In general I prefer not to make any changes in upstream code (including comments), unless we need modification to make it work for Maistra, The reasoning behind is that the more we change upstream code, the harder it becomes to merge and maintain.

That makes sense; so the review in this context ought to be just double checking if the changes reflect what things look like upstream instead of digging and verifying contents, correct? Sorry about the noise in that case.

oschaaf (approved):

dmitri-d:
>  so the review in this context ought to be just double checking if the changes reflect what things look like upstream

Nods, another pair of eyes to make sure the port/cherry-pick didn't go awry. Thanks again!

PR #90, opened 2021-04-28
Fixes MAISTRA-2121: re-enabled handshaker tests

dmitri-d:
This re-enables tls-handshaker tests that were disabled due to issues in test setup, releted to TLS handshake.

### DIFF
--- a/test/extensions/transport_sockets/tls/handshaker_test.cc
+++ b/test/extensions/transport_sockets/tls/handshaker_test.cc
@@ -59,14 +59,19 @@ protected:
     // handshaking.
     auto key = makeKey();
     auto cert = makeCert();
-    auto chain = std::vector<CRYPTO_BUFFER*>{cert.get()};
 
     server_ssl_ = bssl::UniquePtr<SSL>(SSL_new(server_ctx_.get()));
     SSL_set_accept_state(server_ssl_.get());
     ASSERT_NE(key, nullptr);
+    // In TLS1.3 OpenSSL server will send session ticket after an ssl handshake.
+    // While technically not part of the handshake, it's part of the server state-machine,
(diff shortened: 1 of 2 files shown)


test/extensions/transport_sockets/tls/handshaker_test.cc
```diff
@@ -59,14 +59,19 @@ class HandshakerTest : public SslCertsTest {
...
+    // While technically not part of the handshake, it's part of the server state-machine,
+    // and SSL_do_handshake will return errors (SSL_ERROR_WANT_WRITE) on the server-side if 
+    // the session tickets (2 by default) have not been read by the client.
+    // To avoid this altogether, we disable session tickets by setting the number of tickets
+    // to generate for a new session to zero.
+    SSL_set_num_tickets(server_ssl_.get(), 0);
 
-    ASSERT_EQ(1, SSL_set_chain_and_key(server_ssl_.get(), chain.data(), chain.size(), key.get(),
-                                       nullptr));
+    ASSERT_EQ(1, SSL_use_cert_and_key(server_ssl_.get(), cert.get(), key.get(), nullptr, 1));
```
oschaaf, line 74: I personally find it helpful when such parameters have an inline comment, e.g. instead of `nullptr, 1` have something like`/*chain=*/nullptr, /*override=*/1`.

oschaaf (approved):
One style nit/ask, other then that lgtm

test/extensions/transport_sockets/tls/handshaker_test.cc
dmitri-d: I like it on calls with large number of parameters (think SslContext ctor, for example), esp. if they span multiple lines: it's easy to get lost in such cases. I think this call is fairly readable, and something like Visual Studio Code will show parameter list when the cursor is above this line. A lot of SSL api calls have weird parameters, I think adding comments to them will get messy very quickly...

PR #92, opened 2021-05-05
Fixes MAISTRA-2324: Backport multi-root support for Envoy

dmitri-d:
This is a port of tls: separate out cert validation logic from ContextImpl (#14757) (the original PR: https://github.com/envoyproxy/envoy/pull/14757). 

The next PR in the queue is https://github.com/envoyproxy/envoy/pull/14884. It's going to be tricky, upstream relies on what is an OpenSSL internal data-structure in their spiffe validator implementation.

@dgn, @oschaaf.


dmitri-d:
https://github.com/envoyproxy/envoy/pull/14884 ported. 

@dgn, @oschaaf

@oschaaf: fixed the comparison you pointed out the other day; The main changes in this batch are here: https://github.com/maistra/envoy/pull/92/files#diff-067aae04811b0f9ec5191388ebd58efe08012c3bc75e0951c028448ec87e2d70R164. I'm quite certain we are ok (see the comment), an alternative would be to open up internal OpenSSL structures and just use upstream code without changes (something I try to do sparingly).

source/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator.cc
```diff
@@ -0,0 +1,308 @@
...
+    }
+    stats_.fail_verify_error_.inc();
+    return 0;
+  }
+
+  // Set the trust bundle's certificate store on the context, and do the verification.
+  bssl::UniquePtr<X509_STORE_CTX> verify_ctx(X509_STORE_CTX_new());
+  // We make a copy of X509_VERIFY_PARAMs in the store_ctx that we received as a parameter.
+  // This is a precaution mostly, as Envoy doesn't configure any X509_VERIFY_PARAMs.
+  // Note that there's no api to copy crls from one store_ctx to another; the assumption is that
```
oschaaf, line 161: This sentence is slightly hard to grok because of the double negation; if it is semantically correct probably not worth iterating on, but if this is accidental perhaps it is.

oschaaf (approved):
One comment, other then that LGTM

source/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator.cc
dmitri-d: Ugh, indeed, this is hard to understand, and I wrote it. Fixed.

PR #100, opened 2021-07-08
Implement handling of escaped slash characters in URL path (#63)

dmitri-d:
A cherry-pick from istio-envoy-1.9.5. @jwendell: this should resolve integration test failures (here: https://prow.maistra.io/view/gs/maistra-prow-testing/pr-logs/pull/maistra_istio/395/istio_2.1-integration/1413226960250736640).


jwendell:
Why not bump our proxy and envoy to 1.9.5? That's the Istio version we are using.

dmitri-d:
> Why not bump our proxy and envoy to 1.9.5? That's the Istio version we are using.

I don't understand what you mean. With this cherry-pick it *is* 1.9.5 + our changes.

jwendell:
> > Why not bump our proxy and envoy to 1.9.5? That's the Istio version we are using.
> 
> I don't understand what you mean. With this cherry-pick it _is_ 1.9.5 + our changes.

Sorry, I was confused. I thought you had picked only one commit to cherry pick (which turns out to be true, but like you said, this is the only commit).

jwendell (approved):
Thanks.

oschaaf (approved):

PR #113, opened 2021-10-01
[OSSM-553] Enable lua extension on ppc64le

yselkowitz:
Commit Message: Enable lua extension on ppc64le
Additional Description: Now that luajit2 is imported and enabled automatically on ppc, this is needed to actually build the http.lua extension.
Risk Level: Medium
Testing: Built on ppc64le (downstream)
Docs Changes: ??
Release Notes: ??
Platform Specific Features: Affects ppc64le only.

/cc @jwendell

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -5,7 +5,8 @@ load("@envoy_api//bazel:external_deps.bzl", "load_repository_locations")
 load(":repository_locations.bzl", "REPOSITORY_LOCATIONS_SPEC")
 load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
 
-PPC_SKIP_TARGETS = ["envoy.filters.http.lua"]
+# maistra/envoy uses luajit2 on ppc64le so http.lua can be built
+PPC_SKIP_TARGETS = []
 
 WINDOWS_SKIP_TARGETS = [
     "envoy.tracers.dynamic_ot",


jwendell:
LGTM.
@oschaaf ?

bazel/repositories.bzl
```diff
@@ -5,8 +5,6 @@ load("@envoy_api//bazel:external_deps.bzl", "load_repository_locations")
 load(":repository_locations.bzl", "REPOSITORY_LOCATIONS_SPEC")
 load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
 
-PPC_SKIP_TARGETS = ["envoy.filters.http.lua"]
```
oschaaf, line 8: I'm curious, assuming that diverging minimally from upstream is something we want: would initialising to an empty array here work and be preferable? It seems this whole PR could perhaps be reduced to this (plus perhaps a comment to annotate why we diverge here):

```bazel
PPC_SKIP_TARGETS = []
```

oschaaf (commented):
the change makes sense to me, just a question about the specifics on how we pull it off

bazel/repositories.bzl
jwendell: Yeah, that makes sense to me. It should work fine.

oschaaf (approved):

jwendell:
merging manually due to time constraints, agreed with @oschaaf

PR #116, opened 2021-10-21
MAISTRA-2687 Fix spiffe certificate verification

luksa:
Previously, when copying the X509 store context, the code erroneously copied the output certificate chain instead of the input (untrusted) chain to the new context. As the output certificate chain is set only after you run X509_verify_cert(), this chain was always empty. Thus, the code only worked properly when the leaf certificate was signed by one of the certificates in the trust bundle (e.g. when the leaf certificate was signed directly by the root certificate). In cases involving intermediate certificates, the leaf certificate was only trusted if all the intermediate certificates were in the trust bundle. Intermediate certificates sent by the client were ignored.

When copying the X509 store context, the input intermediate certificates should be retrieved by calling X509_STORE_CTX_get0_untrusted(), not X509_STORE_CTX_get0_chain().


oschaaf (commented):
Nice catch! I wonder would it be hard to add a test in test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc?

dgn:
really good catch. I wonder if our copying/re-init of a fresh x509 store is actually necessary?

twghu (commented):
Checking the validity of this change in relation to previous upstream commit.

test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc
```diff
@@ -261,6 +261,27 @@ name: envoy.tls.cert_validator.spiffe
...
+  )EOF"));
+
+  X509StorePtr ssl_ctx = X509_STORE_new();
+
+  // Chain contains workload, intermediate, and ca cert, so it should be accepted.
+  auto cert = readCertFromFile(TestEnvironment::substitute(
+      "{{ test_rundir }}/test/extensions/transport_sockets/tls/test_data/spiffe_san_signed_by_intermediate_cert_chain.pem"));
+  X509StoreContextPtr store_ctx = X509_STORE_CTX_new();
+  EXPECT_TRUE(X509_STORE_CTX_init(store_ctx.get(), ssl_ctx.get(), cert.get(), nullptr));
+  EXPECT_TRUE(validator().doVerifyCertChain(store_ctx.get(), nullptr, *cert, nullptr));
```
oschaaf, line 292: I executed this test on my dev machine, and this expectation doesn't hold there.

oschaaf (changes requested):

test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc
luksa: What's the simplest way for me to run this test on my machine? Is there a container I can use?

oschaaf: Something like this should work (but you'd need to change some paths specific to my dev machine, those are for setting up the bazel build (disk) cache):

```bash
docker run --rm --network=host -it -v "/home/oschaaf/:/config" -v "/home/oschaaf/.cache/bazel/:/home/user/.cache/bazel/" -v "/home/oschaaf/bazel-cache:/bazel-cache"  -v "$(pwd):/tmp/build"  quay.io/maistra-dev/maistra-proxy-builder:2.1  /bin/bash -lc "cd /tmp/build && source /opt/rh/gcc-toolset-9/enable && bazel test  --incompatible_linkopts_to_linklibs --test_output=all  --disk_cache=/bazel-cache //test/extensions/transport_sockets/tls/cert_validator/spiffe/..."
```

oschaaf: (the above line ought to build and run all the tests found under `test/extensions/transport_sockets/tls/cert_validator/spiffe/...`)

twghu:
There appears to be a problem with the spiffe_san_signed_by_intermediate_cert_chain.pem cert, the ca cert included does not match the ca cert using as an anchor in the envoy configuration in the test.  
This fails:
```
openssl verify -verbose -CAfile ca_cert.pem -untrusted  intermediate_ca_cert.pem spiffe_san_signed_by_intermediate_cert.pem
C = US, ST = California, L = San Francisco, O = Lyft, OU = Lyft Engineering, CN = Test Server
error 20 at 0 depth lookup: unable to get local issuer certificate
error spiffe_san_signed_by_intermediate_cert.pem: verification failed
```

If we run the openssl/demos/certs/mkcerts.sh script (openssl source) and then run
```
openssl verify -CAfile root.pem -untrusted intca.pem server.pem
server.pem: OK
```

test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc
luksa: Thanks! This worked like a charm.

luksa:
> There appears to be a problem with the spiffe_san_signed_by_intermediate_cert_chain.pem cert, the ca cert included does not match the ca cert using as an anchor in the envoy configuration in the test. This fails:

Thanks. Yeah, I built the certs by running `certs.sh`. I did verify the new cert with openssl and it worked. But running certs.sh replaces the ca and intermediate cert, so I reverted them before committing. Obviously, the new cert was no longer valid, because the ca cert was now different.

test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc
```diff
@@ -261,6 +261,39 @@ name: envoy.tls.cert_validator.spiffe
...
+  STACK_OF(X509) *intermediates = sk_X509_new_null();
+  sk_X509_push(intermediates, ca_cert.get());
+  sk_X509_push(intermediates, intermediate_ca_cert.get());
+  sk_X509_push(intermediates, cert.get()); // not necessary, but reflects real-world use-case
+
+  X509StoreContextPtr store_ctx = X509_STORE_CTX_new();
+  EXPECT_TRUE(X509_STORE_CTX_init(store_ctx.get(), ssl_ctx.get(), cert.get(), intermediates));
+  EXPECT_TRUE(validator().doVerifyCertChain(store_ctx.get(), nullptr, *cert, nullptr));
+
+  sk_X509_pop_free(intermediates, X509_free);
```
luksa, line 294: Not sure if this gets called when `EXPECT_TRUE` fails. Can't find the definition of `EXPECT_TRUE`.

oschaaf: As far as I know, `sk_X509_pop_free` ought to get called. `ASSERT_TRUE` would stop execution, `EXPECT_TRUE` allows execution to continue.

test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc
oschaaf: Nice. Perhaps we should add this to `/README.md` to make it more discoverable to reduce friction for those who would like to contribute in the future.

oschaaf (approved):
This looks good to me. 
Deferring to @twghu for final review.

test/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator_test.cc
twghu: ASSERT: Fails fast, aborting the current function.
    EXPECT: Continues after the failure.

twghu (approved):

twghu:
Looks good to me.  Builds and tests pass.

jwendell:
Dumb question: Does this affects only our fork? Or upstream too?

oschaaf:
@jwendell good call, so after comparing the upstream code, I think you are right: it looks like this affects upstream as well.

luksa:
@jwendell @oschaaf As far as I can tell, this issue in only in our code, because we clone `store_ctx` (I'm not sure why). Upstream executes `X509_verify_cert(store_ctx)`, whereas we invoke `X509_verify_cert(verify_ctx)`. Given that my fix copies the intermediate certs from `store_ctx` to `verify_ctx`, it's clear that the `store_ctx` contains the certificates. Therefore, the verification should work properly.

twghu:
It would make sense to add the BoringSSL version of this chain validation test to upstream.

dmitri-d:
> As far as I can tell, this issue in only in our code, because we clone store_ctx (I'm not sure why). 

A fly-by comment: BoringSSL exposes store_ctx/context_ctx data structures and makes all their fields directly accessible. OpenSSL hides details of those and only allows access and modification for some of the fields (and only via api), hence all these hoops... Sorry about the bug!

luksa:
@twghu Here's the upstream PR: https://github.com/envoyproxy/envoy/pull/18914

@dmitri-d Thanks for the explanation. No worries about the bug.

dgn (approved):

PR #129, opened 2021-11-22
Adding .gitleaks for identifying test data secrets

twghu:
Signed-off-by: Tim Walsh <temporal.differential@gmail.com>

Red Hat Info Sec identified several embedded secrets in the test data.

This change identifies the files in .gitleaks.toml

Confirmation (test):
```
cd ~/envoy-source-dir
 docker run --rm -v ${PWD}:/my-repo zricethezav/gitleaks:latest --verbose --path="/my-repo"  --repo-config-path=.gitleaks.toml
```
The file paths are fully qualified deliberately as using wild cards (regexps) will potentially allow for missing secret files placed in the directory hierarchy.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,5 +1,9 @@
 title = "Maistra Envoy gitleaks configuration"
 
+# The following rule lists all affected files and folders
+# to account for all known files so any additions can be flagged.
+# To be complete each file should have a regexp to identify
+# the specific key in question (todo). 
 [[rules]]
    description = "PRIVATE KEY CHECK PATHS"
    regex = '''-----BEGIN ((EC|PGP|DSA|RSA|OPENSSH) )?PRIVATE KEY( BLOCK)?-----'''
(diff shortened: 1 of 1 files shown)


.gitleaks.toml
```diff
@@ -0,0 +1,25 @@
+title = "Maistra Envoy gitleaks configuration"
+
+[[rules]]
+   description = "PRIVATE KEY CHECK PATHS"
+   regex = '''-----BEGIN ((EC|PGP|DSA|RSA|OPENSSH) )?PRIVATE KEY( BLOCK)?-----'''
+   tags = ["key"]
+   [rules.allowlist]
+      description = "test files"
+      paths = [
+         '''^test/common/grpc/service_key.json$''',
```
jwendell, line 14: Can we just whitelist everything under test or examples directories?

oschaaf: +1, I agree, intuitively we don't have to worry about keys associated to tests and examples, so perhaps we could condense this list to something like `_test.cc` plus `^(test|examples)/.*`?

oschaaf: @twghu kindly pinging this

twghu: The problem relating to secret disclosure is in part the unintentional inclusion of files or content in files.  Listing each explicitly means we have identified the known problems.  Using a broad "anything in this folder" would allow for unintentional inclusions which by definition is the exact problem we are trying to avoid.

In fact, simply listing all the files is also by definition insufficient.  A complete solution would be to include RegExs for all included secrets in any files or as files.

oschaaf: Thanks for the explanation, with that this LGTM.
Considering that this wasn't obvious to me and John, perhaps that explanation is worth adding as a comment in the `.toml` file?

oschaaf (approved):
LGTM, thanks for following up

PR #140, opened 2022-03-17
[2.2] A few tweaks 

oschaaf:
Signed-off-by: Otto van der Schaaf <ovanders@redhat.com>

### DIFF
--- a/source/extensions/transport_sockets/tls/context_impl.cc
+++ b/source/extensions/transport_sockets/tls/context_impl.cc
@@ -997,12 +997,8 @@ int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
       stats_.ocsp_staple_failed_.inc();
       return SSL_TLSEXT_ERR_ALERT_FATAL;
     }
-    if (SSL_set_tlsext_status_ocsp_resp(ssl, raw_bytes_copy, raw_bytes_size) == 0) {
-      ENVOY_LOG_EVERY_POW_2_MISC(error, "SSL_set_tlsext_status_ocsp_resp failure");
-      OPENSSL_free(raw_bytes_copy);
-      stats_.ocsp_staple_failed_.inc();
-      return SSL_TLSEXT_ERR_ALERT_FATAL;
-    }
+    RELEASE_ASSERT(SSL_set_tlsext_status_ocsp_resp(ssl, raw_bytes_copy, raw_bytes_size),
+                   "SSL_set_tlsext_status_ocsp_resp failure");
     stats_.ocsp_staple_responses_.inc();
   }
     return SSL_TLSEXT_ERR_OK;
--- a/source/extensions/transport_sockets/tls/context_impl.cc
+++ b/source/extensions/transport_sockets/tls/context_impl.cc
@@ -989,10 +989,20 @@ int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
     // We avoid setting the OCSP response if the client didn't request it, but doing so is safe.
     RELEASE_ASSERT(cert_context.ocsp_response_,
                    "OCSP response must be present under OcspStapleAction::Staple");
-    auto& resp_bytes = cert_context.ocsp_response_->rawBytes();
-    int rc = SSL_set_tlsext_status_ocsp_resp(ssl, const_cast<unsigned char*>(resp_bytes.data()),
-                                             resp_bytes.size());
-    RELEASE_ASSERT(rc != 0, "Error setting ocsp response");
+    const std::vector<uint8_t>& raw_bytes = cert_context.ocsp_response_->rawBytes();
+    const std::size_t raw_bytes_size = raw_bytes.size();
+    unsigned char* raw_bytes_copy = static_cast<unsigned char *>(OPENSSL_memdup(raw_bytes.data(), raw_bytes_size));
+    if (raw_bytes_copy == nullptr) { 
+      ENVOY_LOG_EVERY_POW_2_MISC(error, "OPENSSL_memdup failure");
+      stats_.ocsp_staple_failed_.inc();
+      return SSL_TLSEXT_ERR_ALERT_FATAL;
+    }
+    if (SSL_set_tlsext_status_ocsp_resp(ssl, raw_bytes_copy, raw_bytes_size) == 0) {
+      ENVOY_LOG_EVERY_POW_2_MISC(error, "SSL_set_tlsext_status_ocsp_resp failure");
+      OPENSSL_free(raw_bytes_copy);
+      stats_.ocsp_staple_failed_.inc();
+      return SSL_TLSEXT_ERR_ALERT_FATAL;
+    }
     stats_.ocsp_staple_responses_.inc();
   }
     return SSL_TLSEXT_ERR_OK;
--- a/source/extensions/transport_sockets/tls/io_handle_bio.cc
+++ b/source/extensions/transport_sockets/tls/io_handle_bio.cc
@@ -41,11 +41,14 @@ int io_handle_free(BIO* bio) {
     }
     BIO_set_init(bio, 0);
     BIO_clear_flags(bio, INT_MAX);
-    auto* meth = static_cast<BIO_METHOD*>(BIO_get_app_data(bio));
-    if (meth != nullptr) {
-      BIO_meth_free(meth);
-    }
   }
+
+  auto* meth = static_cast<BIO_METHOD*>(BIO_get_app_data(bio));
+  if (meth != nullptr) {
+    BIO_meth_free(meth);
+    BIO_set_app_data(bio, nullptr);
+  }
+
   return 1;
 }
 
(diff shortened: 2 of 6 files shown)


source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -989,10 +989,20 @@ int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
...
+    const std::size_t raw_bytes_size = raw_bytes.size();
+    unsigned char* raw_bytes_copy = static_cast<unsigned char *>(OPENSSL_memdup(raw_bytes.data(), raw_bytes_size));
+    if (raw_bytes_copy == nullptr) { 
+      ENVOY_LOG_EVERY_POW_2_MISC(error, "OPENSSL_memdup failure");
+      stats_.ocsp_staple_failed_.inc();
+      return SSL_TLSEXT_ERR_ALERT_FATAL;
+    }
+    if (SSL_set_tlsext_status_ocsp_resp(ssl, raw_bytes_copy, raw_bytes_size) == 0) {
+      ENVOY_LOG_EVERY_POW_2_MISC(error, "SSL_set_tlsext_status_ocsp_resp failure");
+      OPENSSL_free(raw_bytes_copy);
```
jwendell: It seems you are making a copy of memory, using it and freeing it. What's the issue with the original approach of using the memory directly without making a copy?

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -989,10 +989,20 @@ int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
...
+      stats_.ocsp_staple_failed_.inc();
+      return SSL_TLSEXT_ERR_ALERT_FATAL;
+    }
```
jwendell: You're leaking the copy? Shouldn't it be freed regardless the result of `SSL_set_tlsext_status_ocsp_resp`?

source/extensions/transport_sockets/tls/io_handle_bio.cc
```diff
@@ -123,13 +126,18 @@ long io_handle_ctrl(BIO* b, int cmd, long num, void*) {
 BIO* BIO_new_io_handle(Envoy::Network::IoHandle* io_handle) {
   BIO* b;
   BIO_METHOD* meth = BIO_meth_new(BIO_get_new_index(), "io_handle");
+  bool ok = meth != nullptr;
+
```
jwendell, line 130: does it make sense to fail early if `meth == nullptr`?

source/extensions/transport_sockets/tls/ssl_handshaker.cc
```diff
@@ -159,6 +159,7 @@ const std::string& SslHandshakerImpl::urlEncodedPemEncodedPeerCertificateChain()
         absl::StrReplaceAll(
             pem, {{"\n", "%0A"}, {" ", "%20"}, {"+", "%2B"}, {"/", "%2F"}, {"=", "%3D"}}));
   }
+  sk_X509_pop_free(cert_chain, X509_free);
```
jwendell, line 162: is this fixing a leak? is this present upstream too or it's our addition?

test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
```diff
@@ -522,7 +522,7 @@ TEST_P(SslCertficateIntegrationTest, BothEcdsaAndRsaOnlyRsaOcspResponse) {
 }
 
 // Server has two certificates, but only ECDSA has OCSP, which should be returned.
-TEST_P(SslCertficateIntegrationTest, DISABLED_BothEcdsaAndRsaOnlyEcdsaOcspResponse) {
+TEST_P(SslCertficateIntegrationTest, BothEcdsaAndRsaOnlyEcdsaOcspResponse) {
```
jwendell, line 525: nice :)

source/extensions/transport_sockets/tls/context_impl.cc
oschaaf: So in this flow, we created the copy, but then `SSL_set_tlsext_status_ocsp_resp` failed. As we still own the memory we need to release it. Having said that, it's hard to find examples out there, as everyone seems to ignore the return value (checked httpd,nginx,trafficserver). 
Interestingly httpd does some more sanity checking of what goes into the other args: https://github.com/apache/httpd/blob/ea2c84a0e33d7f2564e77633f8a6c56046fa1618/modules/ssl/ssl_util_stapling.c#L836

source/extensions/transport_sockets/tls/context_impl.cc
oschaaf: IIRC when `SSL_set_tlsext_status_ocsp_resp` succeeds, that also means we transferred ownership of the memory. It's not clearly documented but this is what I remember inferring from reading up on this.

source/extensions/transport_sockets/tls/ssl_handshaker.cc
oschaaf: I haven't checked, but it's probably specific to us. Pretty sure that upstream this  would have been caught it in CI runs

source/extensions/transport_sockets/tls/io_handle_bio.cc
oschaaf: Well I tried to simplify control flow and maximize readability while addressing all the possible ways this code may fail in this code, by conditioning everything on the status we track in `ok`. Otherwise we'd get lots of early returns and if statements.. so I am leaning to keep this as is?

source/extensions/transport_sockets/tls/context_impl.cc
jwendell: It looks like it never fails anyway. It just transfers onwership: https://github.com/openssl/openssl/blob/master/ssl/s3_lib.c#L3552-L3556
So, we're good.

source/extensions/transport_sockets/tls/io_handle_bio.cc
jwendell: Yeah, makes sense. I was just wondering if it's worth to execute the whole function if `meth == nullptr`. It looks like they don't check for null args, which may lead to a crash: https://github.com/openssl/openssl/blob/master/crypto/bio/bio_meth.c#L89

jwendell (approved):

source/extensions/transport_sockets/tls/io_handle_bio.cc
oschaaf: resolving (John and me discussed offline)

source/extensions/transport_sockets/tls/context_impl.cc
oschaaf: re: it never fails anyway
Right - thanks for pointing that out. So reading that, I think a more pragmatic approach then is to not go and attempt to handle this situation gracefully, but instead have a `RELEASE_ASSERT`. 
That way we'll learn about it soon when this situation changes (theoretically there's nothing that says the call couldn't start doing it in future versions), while avoiding code that for practical purposes should be considered dead.

source/extensions/transport_sockets/tls/openssl_impl.cc
```diff
@@ -49,14 +49,11 @@ int set_strict_cipher_list(SSL_CTX* ctx, const char* str) {
...
   STACK_OF(X509)* to_copy = SSL_get_peer_cert_chain(ssl);
-  if (!to_copy) {
-    return nullptr;
-  }
-  STACK_OF(X509)* ret = sk_X509_dup(SSL_get_peer_cert_chain(ssl));
-
-  if (SSL_is_server(ssl)) {
+  // sk_X509_dup does not increase reference counts on certs in the stack.
+  STACK_OF(X509)* ret = to_copy == nullptr ? nullptr : sk_X509_dup(to_copy);
+  if (ret != nullptr && SSL_is_server(ssl)) {
```
tedjpoole, line 56: I can't see what this is fixing? It doesn't seem to change any behaviour other than eliminate a duplicate/redundant call to SSL_get_peer_cert_chain()

oschaaf: Hmm, I don't clearly recall unfortunately, but I think that eliminating the duplicate call was the sole goal here. So that is what it addresses. I think I did this in an attempt to simplify reasoning while debugging memory issues.

tedjpoole (approved):

source/extensions/transport_sockets/tls/openssl_impl.h
```diff
@@ -19,7 +19,13 @@ namespace Tls {
 
 int set_strict_cipher_list(SSL_CTX* ctx, const char* str);
 
-STACK_OF(X509) * SSL_get_peer_full_cert_chain(const SSL* ssl);
+// SSL_get_peer_full_cert_chain exists in the BoringSSL library. This diverges in that
+// the caller will own the returned certificate chain and needs to call sk_X509_free on any
```
oschaaf: Whoops it seems mentioning `sk_X509_free()` here is inaccurate, that ought to be `sk_X509_pop_free()`

oschaaf: Addressed in c3b2e8e5d6926d33cd31341ebf82c47670841159

source/extensions/transport_sockets/tls/context_impl.cc
oschaaf: Addressed in c3b2e8e5d6926d33cd31341ebf82c47670841159 (added a `RELEASE_ASSERT` and removed the code that attempted to handle this)

source/extensions/transport_sockets/tls/context_impl.cc
oschaaf: I chased our handling of the ocsp responses, and we don't need to add further sanity checking. Envoy's handling is on par with httpd in terms of sanity checking.

twghu (approved):

PR #141, opened 2022-03-18
Fixes to build proxy

jwendell:


WORKSPACE
```diff
@@ -29,3 +29,8 @@ new_local_repository(
     build_file = "openssl.BUILD",
     path = "/usr/lib64/",
 )
+new_local_repository(
```
oschaaf, line 32: nit: add newline (though perhaps we should start using Envoy's fix-format tool and enforce that in CI?).

contrib/sip_proxy/filters/network/source/router/router_impl.h
```diff
@@ -241,7 +241,7 @@ class TransactionInfo : public std::enable_shared_from_this<TransactionInfo>,
   std::shared_ptr<UpstreamRequest> getUpstreamRequest(const std::string& host) {
     try {
       return tls_->getTyped<ThreadLocalTransactionInfo>().upstream_request_map_.at(host);
-    } catch (std::out_of_range) {
+    } catch (std::out_of_range const&) {
```
oschaaf, line 244: This would be a great candidate for changing upstream too?

bazel/dependency_imports.bzl
```diff
@@ -11,7 +11,7 @@ load("@proxy_wasm_rust_sdk//bazel:dependencies.bzl", "proxy_wasm_rust_sdk_depend
 load("@rules_cc//cc:repositories.bzl", "rules_cc_dependencies", "rules_cc_toolchains")
 
 # go version for rules_go
-GO_VERSION = "1.15.5"
+GO_VERSION = "host"
```
oschaaf, line 14: Can we add a one line doc comment to explain the rationale for this change?

bazel/foreign_cc/BUILD
```diff
@@ -253,106 +253,6 @@ envoy_cmake(
     }),
 )
 
-envoy_cmake(
```
oschaaf, line 256: perhaps a one line comment about why we don't need this would be helpful?

Also - I wonder - can we figure out a way to minimize this change? E.g. for windows this has `tags = ["skip_on_windows"]`, maybe we would benefit from something similar. In that case this change would fold to a single line + a comment on why we do that. This might make life easier when cutting a branch for a major new release (when we need to rebase/merge)? Just some thoughts.

oschaaf (approved):
This is awesome, and LGTM overall.
I left some comments/nits/asks; I will leave it up to you to decide if you want to follow up, and if so, if you want to do this in this PR or mop them up via a follow up.
Approving, but I added a label to keep it on hold to give you an opportunity to make a decision with respect to the review comments I left.

WORKSPACE
jwendell: +1 for enforcing this in CI.

contrib/sip_proxy/filters/network/source/router/router_impl.h
jwendell: This *is* from upstream, it's a cherry-pick. It is a separate commit in this PR.

bazel/dependency_imports.bzl
jwendell: This is a port from https://github.com/maistra/proxy/blob/maistra-2.1/maistra/scripts/update-deps.sh#L97 - I removed that patch from proxy in 2.2.
We use the host's golang (as we use java, gcc, etc from the host)

bazel/foreign_cc/BUILD
jwendell: This - and other changes in this PR - got lost in the rebase. We did that in previous versions. The reason I did that this way back in 2.1 or 2.0? Lack of knowledge in Bazel. I'd love to simplify things where it's possible. Let's do this in a follow up though.

contrib/sip_proxy/filters/network/source/router/router_impl.h
oschaaf: ah I see, sorry, missed that.

PR #142, opened 2022-03-22
Ensure librdkafka configured correctly if compiler warnings are disabled

tedjpoole:
When building envoy in the context of maistra/proxy, many compiler warnings are disabled, which causes librdkafka's configure step to incorrectly assume that strlcpy() exists, when in fact it doesn't. This causes librdkafka to get built with undefined references to strlcpy() which then causes the envoy binary link step to fail.

This change ensures that librdkafka gets configured correctly regardless of what compiler warnings are in effect at the higher level

### DIFF
--- a/bazel/foreign_cc/BUILD
+++ b/bazel/foreign_cc/BUILD
@@ -44,7 +44,7 @@ cc_library(
 configure_make(
     name = "librdkafka_build",
     configure_in_place = True,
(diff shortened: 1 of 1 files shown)


jwendell (changes requested):
This is not an issue for us. We are not disabling warnings in 2.2 anymore.

oschaaf (approved):
lgtm, defering to @jwendell for final review.

jwendell (approved):

PR #146, opened 2022-03-24
OSSM-1264: Fix issues that prevent changes to be propagated to proxy

jwendell:


oschaaf (approved):
As discussed on Slack I would appreciate leaving a breadcrumb in the form of a doc comment to explain the rationale behind this patch. A hypothetical future person running a 3-way comparison can use all the help he can get to reconstruct context, and I'm sure he/she may thank us for it.
Having said that, up to you, LGTM

bazel/repositories.bzl
```diff
@@ -722,7 +722,11 @@ def _com_google_protobuf():
 
     external_http_archive(
         "com_google_protobuf",
-        patches = ["@envoy//bazel:protobuf.patch"],
+        patches = [
+            "@envoy//bazel:protobuf.patch",
+            # This patch adds the protobuf_version.bzl file to the protobuf tree, which is missing from the 3.18.0 tarball.
```
oschaaf, line 727: that begs the question: where did you get the right `protobuf_version.bzl` from? 
is this something that needs to be repeated as the protobuf dep version gets bumped?

jwendell: this is fixed in later protobuf. I believe this issue is only present on 3.18.x. Get a 3.18 tarball and extract it. This file is not there: https://github.com/protocolbuffers/protobuf/tags - However, the file is present in the 3.18 branch. We could propose this patch in envoy 1.20 but I think nobody cares.

oschaaf: I see, thank you for explaining, makes sense.

PR #150, opened 2022-03-28
Fixed ownership of SSL_set_tlsext_status_ocsp_resp() parameter

tedjpoole:
The SSL_set_tlsext_status_ocsp_resp() function takes ownership of the OCSP response buffer parameter, but it was being passed the underlying buffer owned by a vector<>. This resulted in (1) an attempt to OPENSSL_free() the buffer that had been allocated by new, and (2) the vector<> attempting to free it's buffer, which had already been free'd.

Fixed by adding an OPENSSL_memdup() on the buffer.

### DIFF
--- a/test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
+++ b/test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
@@ -522,7 +522,7 @@ TEST_P(SslCertficateIntegrationTest, BothEcdsaAndRsaOnlyRsaOcspResponse) {
 }
 
 // Server has two certificates, but only ECDSA has OCSP, which should be returned.
-TEST_P(SslCertficateIntegrationTest, DISABLED_BothEcdsaAndRsaOnlyEcdsaOcspResponse) {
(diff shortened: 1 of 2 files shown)


oschaaf:
Doesn’t https://github.com/maistra/envoy/pull/140 incorporate this?

test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
```diff
@@ -522,7 +522,7 @@ TEST_P(SslCertficateIntegrationTest, BothEcdsaAndRsaOnlyRsaOcspResponse) {
 }
 
 // Server has two certificates, but only ECDSA has OCSP, which should be returned.
-TEST_P(SslCertficateIntegrationTest, DISABLED_BothEcdsaAndRsaOnlyEcdsaOcspResponse) {
+TEST_P(SslCertficateIntegrationTest, BothEcdsaAndRsaOnlyEcdsaOcspResponse) {
```
oschaaf, line 525: Question: did all three tests fail without the above changes?

oschaaf (approved):
modulo the changes we just discussed out of band around error handling, this lgtm, awesome.

tedjpoole:
Closing because the same fix is in #140

PR #154, opened 2022-04-05
OSSM-1063: Switch to clang

jwendell:


bazel/envoy_internal.bzl
```diff
@@ -7,7 +7,7 @@ def envoy_copts(repository, test = False):
     posix_options = [
         "-Wall",
         "-Wextra",
-        "-Werror",
+#        "-Werror", FIXME: https://issues.redhat.com/browse/OSSM-1201
```
oschaaf, line 10: We could remove this line?

bazel/foreign_cc/luajit.patch
```diff
@@ -11,15 +11,6 @@ index e65b55e..f0a61dd 100644
  #
  # Use this if you want to force a 32 bit build on a 64 bit multilib OS.
  #CC= $(DEFAULT_CC) -m32
-@@ -36,7 +36,7 @@ CC ?= $(DEFAULT_CC)
- # to slow down the C part by not omitting it. Debugging, tracebacks and
- # unwinding are not affected -- the assembler part has frame unwind
- # information and GCC emits it where needed (x64) or with -g (see CCDEBUG).
--CCOPT= -O2 -fomit-frame-pointer
-+CCOPT= -O2 -fomit-frame-pointer -fPIC -fPIE -fplugin=annobin -fstack-protector-strong -fstack-protector-all
```
oschaaf, line 19: question: this looks like annobin & at least some of the hardening flags work with annobin?

bazel/repositories.bzl
```diff
@@ -821,10 +821,7 @@ def _com_googlesource_chromium_v8():
         name = "com_googlesource_chromium_v8",
         genrule_cmd_file = "@envoy//bazel/external:wee8.genrule_cmd",
         build_file = "@envoy//bazel/external:wee8.BUILD",
-        patches = [
-            "@envoy//bazel/external:wee8.patch",
-            "@envoy//bazel/external:wee8-annobin.patch",
```
oschaaf, line 826: is the plan to re-enable this patch for annobin in a follow up?

maistra/bazelrc
```diff
@@ -0,0 +1,5 @@
+# FIXME: https://issues.redhat.com/browse/OSSM-1201
+build --cxxopt -w
```
oschaaf, line 2: Do we still need this? (No objections to merging as-is and mopping up later if so, just curious)

maistra/run-ci.sh
```diff
@@ -16,13 +16,12 @@ export BUILD_SCM_REVISION="Maistra PR #${PULL_NUMBER:-undefined}"
 export BUILD_SCM_STATUS="SHA=${PULL_PULL_SHA:-undefined}"
 
 COMMON_FLAGS="\
-    --incompatible_linkopts_to_linklibs \
+    --config=clang \
```
oschaaf, line 19: I wonder, can we do without this ` --config=clang`?

oschaaf (approved):
LGTM, awesome. Left a few questions, but nothing blocking.

bazel/envoy_internal.bzl
jwendell: We could, but I slightly prefer to keep those around, as they keep bothering us, thus motivating us to not forget about them.

bazel/foreign_cc/luajit.patch
jwendell: yeah, for now I'm removing everything. We can work towards re-enabling them with the possible exception of annobin.

bazel/repositories.bzl
jwendell: Kind of. The plan is to investigate which options are necessary. If we figure out we don't need any of them, we should get rid of this patch permanently.

maistra/bazelrc
jwendell: We do. I had a hard time trying to workaround some warnings/errors (-Werror=xxx) in the proxy build. Adding back this flag (we had it in 2.1) is the way to move on with it.

maistra/run-ci.sh
jwendell: Apparently there's a bazel build that does not identify correctly the compiler, which makes bazel/envoy think it's gcc and then enabling gcc-specific options. Then we get *lots* of warnings about unknown flags in clang. This basically declares the variable `BAZEL_COMPILER=clang`

oschaaf: Aaah I see, sounds like there's good a good reason for it to be there 👍

PR #155, opened 2022-04-07
Missing bits from maistra-2.1

tedjpoole:


source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,14 +228,20 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
...
-    uint32_t protocol_length = 0;
-    protocol_length <<= 8;
-    protocol_length |= data[i];
-    ++i;
-    absl::string_view protocol(str.substr(i, protocol_length));
-    protocols.push_back(protocol);
+    uint8_t protocol_length = data[i++];
+
+    if (protocol_length == 0) {
+      ENVOY_LOG(debug, "tls inspector: ALPN protocol length is zero");
```
oschaaf: I wonder if emitting a warning in combination with a switch to `ENVOY_LOG_PERIODIC_TO_LOGGER` [1] would be better here. wdyt? 

[1] https://github.com/envoyproxy/envoy/blob/e727e1f50b33730a3a4fc061f95bcdd6d0c48768/source/common/common/logger.h#L541

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,14 +228,20 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
...
 
-  absl::string_view str(reinterpret_cast<const char*>(data));
   for (unsigned int i = 0; i < len;) {
-    uint32_t protocol_length = 0;
-    protocol_length <<= 8;
-    protocol_length |= data[i];
-    ++i;
-    absl::string_view protocol(str.substr(i, protocol_length));
-    protocols.push_back(protocol);
+    uint8_t protocol_length = data[i++];
```
oschaaf, line 237: So the format is that data is prefixed with a single byte, which represents the number of data bytes, correct? Can you add code-level comment to the method to document Filter::getAlpnProtocols while we are at it?

Tangentially related, but imho out of scope for now -- I'd love it if we could refactor this method into it's own unit of compilation under `/maistra/source/tls/alpn_util.cc` or some such.. and then we could have unit tests for this in `/maistra/source/tls/test_alpn_util.cc`.  That would prove correctness, tease out issues when run under asan, as well as serve as semi-documentation on assumptions and edge cases etc.

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,14 +228,20 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
     return protocols;
   }
 
-  absl::string_view str(reinterpret_cast<const char*>(data));
   for (unsigned int i = 0; i < len;) {
```
oschaaf: should we put a cap on the value of len that we are willing to accept?

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,14 +228,20 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
...
-    protocols.push_back(protocol);
+    uint8_t protocol_length = data[i++];
+
+    if (protocol_length == 0) {
+      ENVOY_LOG(debug, "tls inspector: ALPN protocol length is zero");
+      break;
+    }
+
+    if (protocol_length > (len - i)) {
+      ENVOY_LOG(debug, "tls inspector: ALPN protocol length is too long");
```
oschaaf: ditto

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,14 +228,20 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
...
+      break;
+    }
+
+    protocols.emplace_back(reinterpret_cast<const char*>(data + i), protocol_length);
     i += protocol_length;
```
oschaaf, line 250: readability nit: afaict this increment could move into the usual position in the for() loop on the premise that we declare it outside of that and initialize it to zero -- but that probably is just personal preference

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -737,10 +737,6 @@ ServerContextImpl::ServerContextImpl(Stats::Scope& scope,
       if (!response->matchesCertificate(*tls_context_.cert_contexts_[i].cert_chain_)) {
         throw EnvoyException("OCSP response does not match its TLS certificate");
       }
-      if (response->isExpired()) {
```
oschaaf: Did we verify upstream also stripped this? Or is this specific to us?

source/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator.cc
```diff
@@ -165,9 +165,9 @@ int SPIFFEValidator::doVerifyCertChain(X509_STORE_CTX* store_ctx,
   bssl::UniquePtr<X509_STORE_CTX> verify_ctx(X509_STORE_CTX_new());
   // We make a copy of X509_VERIFY_PARAMs in the store_ctx that we received as a parameter.
   // This is a precaution mostly, as Envoy doesn't configure any X509_VERIFY_PARAMs.
-  // Note that there's no api to copy crls from one store_ctx to another; the assumption is that
-  // neither X509_V_FLAG_CRL_CHECK nor X509_V_FLAG_CRL_CHECK_ALL verify_params are not used.
-  // Should this change, consider opening up X509_STORE_CTX struct, which is internal atm.
+  // Note that there's no api to copy crls from one store_ctx to another; the assumption is that
```
oschaaf, line 168: Are we sure we included the rest of Marko his fix for this validation?

oschaaf (changes requested):
Nice work, like how readable `getAlpnProtocols` got. Left a few nits/questions/comments.

bazel/foreign_cc/luajit2.patch
```diff
@@ -0,0 +1,98 @@
...
++CC ?= $(DEFAULT_CC)
+ #
+ # Use this if you want to force a 32 bit build on a 64 bit multilib OS.
+ #CC= $(DEFAULT_CC) -m32
+@@ -36,7 +36,7 @@ CC= $(DEFAULT_CC)
+ # to slow down the C part by not omitting it. Debugging, tracebacks and
+ # unwinding are not affected -- the assembler part has frame unwind
+ # information and GCC emits it where needed (x64) or with -g (see CCDEBUG).
+-CCOPT= -O2 -fomit-frame-pointer
++CCOPT= -O2 -fomit-frame-pointer -fPIC -fPIE -fplugin=annobin -fstack-protector-strong -fstack-protector-all
```
jwendell: We removed this in a recent pr

bazel/repositories.bzl
```diff
@@ -1006,6 +1006,9 @@ def _com_github_luajit2_luajit2():
     external_http_archive(
         name = "com_github_luajit2_luajit2",
         build_file_content = BUILD_ALL_CONTENT,
+        patches = ["@envoy//bazel/foreign_cc:luajit2.patch"],
```
jwendell, line 1009: Are you sure? I think this patch is already present on 2.2

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: I wasn't too sure about logging level to use. I had a look around and it seemed to me like warning/error log messages are generally used to report on the (mal)functioning of envoy itself, rather than to report on erroneous content flowing through envoy, but I'll change it as suggested.

bazel/repositories.bzl
knm3000: It is currently not present in 2.2: some luajit2 changes have not been merged from 2.1 into 2.2. This PR contains all missing luajit2 code that is needed to build envoy/proxy with luajit2.

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: At the point where getAlpnProtocols() is called the buffer has already been allocated and populated off the wire, so why would we reject what's in it?

oschaaf: makes sense resolving

source/extensions/transport_sockets/tls/context_impl.cc
tedjpoole: This is specific to us and the decision to remove it was made in discussion with @twghu

source/extensions/transport_sockets/tls/cert_validator/spiffe/spiffe_validator.cc
tedjpoole: Yes, it was only the change to the comment that didn't make it to 2.2

source/extensions/transport_sockets/tls/context_impl.cc
oschaaf: great, thanks for sharing context. perhaps leave note for our future selves in the code? e.g. something like:

```
// MAISTRA - removed "if (response->isExpired()" handling, because .. 
```

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
oschaaf: The thing is that with regular logging this could result in far too much log spew. But i think that hitting these edge cases is worth surfacing to the user. With the periodic log function, we don't be flooding the logs while still surfacing that this stuff is happening. That was my reasoning.

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,15 +228,25 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
...
-    protocol_length |= data[i];
-    ++i;
-    absl::string_view protocol(str.substr(i, protocol_length));
-    protocols.push_back(protocol);
-    i += protocol_length;
+  // The data buffer contains a sequence of non-terminated strings, where each
+  // string is preceded by 1 byte that gives it's length (excluding the length
+  // byte itself). No string can be zero length, and the last string must not
+  // overrun or underun the buffer.
+  uint8_t protocol_length;
```
oschaaf: shall we init protocol_length to 0? :-)

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
```diff
@@ -228,15 +228,25 @@ std::vector<absl::string_view> Filter::getAlpnProtocols(const unsigned char* dat
...
-    protocol_length <<= 8;
-    protocol_length |= data[i];
-    ++i;
-    absl::string_view protocol(str.substr(i, protocol_length));
-    protocols.push_back(protocol);
-    i += protocol_length;
+  // The data buffer contains a sequence of non-terminated strings, where each
+  // string is preceded by 1 byte that gives it's length (excluding the length
+  // byte itself). No string can be zero length, and the last string must not
+  // overrun or underun the buffer.
```
oschaaf: perfect, thank you for adding this. 
just one nit: s/underun/underrun/

oschaaf (changes requested):

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: It didn't look necessary to me because it does get initialised before use.

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: Done

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
oschaaf: Sorry if I'm a little dense, but where does protocol_length get initialised? 
(From what I can tell `i += protocol_length` is the first statement that mentions it?)

tedjpoole: The `i += protocol_length` statement doesn't get executed until after the first run through the loop, so the `protocol_length = data[i++]` statement is the first to be executed.

jwendell: evil :D

oschaaf: Aaah yes, I missed that; thanks for explaining. Resolving this..

oschaaf: though thinking about it, perhaps a one line comment about this won't hurt here just to make it obvious to future readers?

jwendell: or just initialize it... compiler might optimize it anyway, so...

We are also doing other evil stuff like changing the for loop variable in the middle of the loop. It might work, but my personal opinion is that in such cases we could just use while instead of for.

oschaaf: So I mulled that over a bit, and now I agree on that I'd prefer a while loop as well. It seems like we're going through hoops to try and fit what we are doing into the original for loop construction, which I think started us off on the wrong foot: a while loop seems to be a more natural intuitive candidate here. @tedjpoole sorry about the flipflopping :-) curious to hear what you think?

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: Done

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: done

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
tedjpoole: Added comment to explain the format of the data being unmarshalled.

source/extensions/transport_sockets/tls/context_impl.cc
tedjpoole: Reinstated this check because it looks like it has some value. What do you think @twghu? Should I also add it upstream?

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
oschaaf: 👍 thank you, resolving.

bazel/foreign_cc/luajit2.patch
jwendell: This change needs to be removed otherwise it will fail and we will only notice it on ppc builds, not on prow (x86). This is an issue: We currently don't have a way to try builds on non-x86 platforms before the pr is merged.

source/extensions/filters/listener/tls_inspector/tls_inspector.cc
oschaaf: thanks for following up, I like the while loop better. Resolving
/cc @jwendell

oschaaf (approved):
modulo pending comments from others, this LGTM

bazel/repositories.bzl
jwendell: True. I just hit this in a ppc product build. The build fails because of this.

tedjpoole: @jwendell just to be clear, do you mean that you want this left in?

jwendell: Yes. This is missing in 2.2 indeed.

You just need to get rid of the change to the flags mentioned above
```diff
-CCOPT= -O2 -fomit-frame-pointer
+CCOPT= -O2 -fomit-frame-pointer -fPIC -fPIE -fplugin=annobin -fstack-protector-strong -fstack-protector-all
```

bazel/foreign_cc/luajit2.patch
tedjpoole: Done. It builds & tests ok (on x86) using the `--//source/extensions/filters/common/lua:luajit2=1` option.

jwendell: Awesome, thanks.

jwendell (approved):

source/extensions/transport_sockets/tls/context_impl.cc
twghu: The expiry check if() was added 
on 13/4/21 https://github.com/maistra/envoy/commit/412c091b302cc688bf3269022bf679bced828b2f#diff-2b39f212c2f760ae547351a46f92ad08c0ec84c556d678ebe9388b9d008e7313

and was removed on 20/5/21
https://github.com/maistra/envoy/commit/963afc8094dc2c17bd24526f2de11841a5904b1b

The last commit should indicate reason for removal, a sanity check would be a good idea.

oschaaf: So digging into this, it seems the check is (also) done in [ServerContextImpl::ocspStapleAction](https://github.com/maistra/envoy/blob/0428980c9210b2c7de98eb64e73fcfc526c96426/source/extensions/transport_sockets/tls/context_impl.cc#L948).  So I think we can remove it.

PR #139, opened 2022-03-10
MAISTRA-2648: WASM fix for s390x [maistra-2.2]

knm3000:
Fixes https://issues.redhat.com/browse/MAISTRA-2648
The permanent fix is not a part of maistra/envoy yet:
proxy-wasm/proxy-wasm-cpp-host#198

Signed-off-by: Konstantin Maksimov <konstantin.maksimov@ibm.com>

PR #156, opened 2022-04-11
Revert "Revert "MAISTRA-2648: WASM fix for s390x (#139)" (#149)"

oschaaf:
This reverts commit d5e17b5936f4bf166a775d549037a2b248a874de.

This erroneously was reverted while debugging the problems with
Istio's integrationt tests. I think we should undo that.

Signed-off-by: Otto van der Schaaf <ovanders@redhat.com>


jwendell (commented):
This broke 2.1 so we assumed it also broke 2.2. It got reverted in 2.2 but not because the istio test failures, but because the wasm failures on s390x.

oschaaf:
You are right, my recollection failed me. Closing this, sorry about the noise.

PR #159, opened 2022-04-26
Updated zlib to version 1.2.12

tedjpoole:


bazel/repository_locations.bzl
```diff
@@ -429,15 +429,12 @@ REPOSITORY_LOCATIONS_SPEC = dict(
         project_name = "zlib",
         project_desc = "zlib compression library",
         project_url = "https://zlib.net",
-        version = "79baebe50e4d6b73ae1f8b603f0ef41300110aa3",
-        # Use the dev branch of zlib to resolve fuzz bugs and out of bound
-        # errors resulting in crashes in zlib 1.2.11.
-        # TODO(asraa): Remove when zlib > 1.2.11 is released.
-        sha256 = "155a8f8c1a753fb05b16a1b0cc0a0a9f61a78e245f9e0da483d13043b3bcbf2e",
+        version = "21767c654d31d2dccdde4330529775c6c5fd5389",
```
jwendell, line 432: shouldn't you use `1.2.12` here?

jwendell (commented):
Is this necessary? It's building fine as of now, no?

Rather, I'd submit this in upstream, main branch instead of our fork.

tedjpoole:
I have also submitted it upstream, so feel free to discard this PR if you don't want the extra risk of changing maistra now

jwendell:
yeah, I don't think we should change maistra now. It's not an important change as it was in the jwt repo.

PR #153, opened 2022-04-01
OSSM-1297: Use a remote build cache, if it exists

jwendell:
This is intended to work in CI and in local development workflow as
well.

PR #160, opened 2022-04-26
OSSM-1297: Use a remote build cache, if it exists

jwendell:
Manual backport of https://github.com/maistra/envoy/pull/153

### DIFF
--- a/maistra/run-ci.sh
+++ b/maistra/run-ci.sh
@@ -15,13 +15,23 @@ export ARCH
 export BUILD_SCM_REVISION="Maistra PR #${PULL_NUMBER:-undefined}"
 export BUILD_SCM_STATUS="SHA=${PULL_PULL_SHA:-undefined}"
 
+COMMON_FLAGS="\
+    --incompatible_linkopts_to_linklibs \
+    --local_ram_resources=12288 \
+    --local_cpu_resources=6 \
+    --jobs=3 \
+    --color=no \
+"
+
+if [ -n "${BAZEL_REMOTE_CACHE}" ]; then
+  COMMON_FLAGS+=" --remote_cache=${BAZEL_REMOTE_CACHE} "
+elif [ -n "${BAZEL_DISK_CACHE}" ]; then
+  COMMON_FLAGS+=" --disk_cache=${BAZEL_DISK_CACHE} "
+fi
+
 # Build
 time bazel build \
-  --incompatible_linkopts_to_linklibs \
-  --local_ram_resources=12288 \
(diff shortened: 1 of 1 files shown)


maistra/run-ci.sh
```diff
@@ -15,26 +15,32 @@ export ARCH
...
+    --local_cpu_resources=6 \
+    --jobs=3 \
+    --color=no \
+"
+
+if [ -n "${BAZEL_REMOTE_CACHE}" ]; then
+  COMMON_FLAGS+=" --remote_cache=${BAZEL_REMOTE_CACHE} "
+elif [ -n "${BAZEL_DISK_CACHE}" ]; then
+  COMMON_FLAGS+=" --disk_cache=${BAZEL_DISK_CACHE} "
+fi
```
tedjpoole, line 30: Should there be an `else` clause here so the `--disk_cache=/bazel-cache` option still gets used as the default if neither of the environment variables are set?

jwendell: I don't think so. We should set the `BAZEL_DISK_CACHE` variable locally if we want to use it. Not that this is already in place for 2.2. This is just a backport to 2.1.

tedjpoole: I was just thinking about folks who build in a docker container with a `/bazel-cache' mount, and expect it to just work. With this change, the default behaviour will change to using ~/.cache/bazel instead. However, I guess it's simple enough to set the additional environment variable.

tedjpoole (approved):

maistra/run-ci.sh
jwendell: yeah, that's what I'm doing

maistra/run-ci.sh
```diff
@@ -15,26 +15,32 @@ export ARCH
 export BUILD_SCM_REVISION="Maistra PR #${PULL_NUMBER:-undefined}"
 export BUILD_SCM_STATUS="SHA=${PULL_PULL_SHA:-undefined}"
 
+COMMON_FLAGS="\
+    --incompatible_linkopts_to_linklibs \
+    --local_ram_resources=12288 \
+    --local_cpu_resources=6 \
+    --jobs=3 \
```
dgn, line 22: why the reduction to 3 jobs?

dgn (approved):

maistra/run-ci.sh
jwendell: we had issues with 4 jobs in 2.2, so, let's reduce everywhere on demand

PR #169, opened 2022-06-13
OSSM-1344: Fixing issues for ppc64le

NishikantThorat:
- Added changes for building envoy on ppc64le. (Kept single patch file for P & Z)


bazel/repositories.bzl
```diff
@@ -829,7 +829,7 @@ def _com_googlesource_chromium_v8():
         build_file = "@envoy//bazel/external:wee8.BUILD",
         patches = [
             "@envoy//bazel/external:wee8.patch",
-            "@envoy//bazel/external:wee8-s390x.patch",
+            "@envoy//bazel/external:wee8-s390x-ppc64.patch",
```
oschaaf, line 832: It looks like we rename here, but it looks like the patch file is introduces as a new file instead of as a change to the pre-existing one. Maybe use `git mv wee8-s390x.patch wee8-s390x-ppc64.patch` to implement this change as a rename to the pre-existing s390 patch? If not, should we delete the pre-existing patch to clean up (we no longer need it)?

NishikantThorat: "wee8-s390x.patch" added(owned) by Z, so I haven't renamed and added new for P & Z.

jwendell: Please delete the Z one

NishikantThorat: Done

oschaaf (approved):

PR #170, opened 2022-06-13
[2.1] Fix leaking BIO_METHOD objects

tedjpoole:


oschaaf:
This seems like a subset of https://github.com/maistra/envoy/pull/140

Not trying to complicate this, but did we have a look at attempting to see if there's more that would be useful to include in  2.1?

tedjpoole:
Yes, this is a subset of #140. I was just concentrating on OSSM-1593.

Once OSSM-1593 is dealt with, I'll go through the rest of #140 and backport whatever bits are relevant for 2.1

twghu (approved):
LGTM!

PR #180, opened 2022-06-22
tls: remove deprecated cipher suites from client defaults

jforce:
This aligns the maistra fork of envoy to upstream by removing deprecated cipher suites which should not be present as client defaults. 

Please see:
- [PR 13850](https://github.com/envoyproxy/envoy/pull/13850)
- [Issue 5397](https://github.com/envoyproxy/envoy/issues/5397)
- [Issue 5396](https://github.com/envoyproxy/envoy/issues/5396)


oschaaf:
Did you test this change? And a naive question: should we consider cherry-picking upstream's https://github.com/envoyproxy/envoy/commit/06f63510c4b5eff7b61d703674fec26b3b1bbde2 instead?
Digging into this for a bit it seems there's a comment about a change like this over at: https://github.com/maistra/envoy/blob/61a066d1a608ced13d29859b5f4fc0e988e71697/test/extensions/transport_sockets/tls/ssl_socket_test.cc#L4380

jforce:
No I've not tested this

Yeah that's fine I'm not particularly fussed on how we remove these weak ciphers, I just want them gone by default ;) 

I assume that comment is just for the server side of things under `ServerContextConfigImpl` and not the client side

jforce:
Issue to track this change: https://issues.redhat.com/browse/OSSM-1667

twghu:
We have had some discussion regarding making deprecation optional however this would be a risk, consequently we won't be making it optional.

This PR needs an  additional change to ssl_socket_test.cc in maistra 2,2 
specifically:

 // TODO (dmitri-d) Enable this check if we start deprecating cipher-suites.
  // Client connects to a server offering only deprecated cipher suites, connection fails.
  // server_params->add_cipher_suites("ECDHE-RSA-AES128-SHA");
  // updateFilterChain(tls_context, *filter_chain);
  // error_test_options.setExpectedServerStats("ssl.connection_error");
  // testUtilV2(error_test_options);
  server_params->clear_cipher_suites();
  
  If these lines are un-commented then the ssl_socket_test.cc test code will be in sync with upstream.
 I am not sure that there is sufficient testing of deprecations here but at least we will be in sync.
  
 Regarding cherry-picking https://github.com/envoyproxy/envoy/commit/06f63510c4b5eff7b61d703674fec26b3b1bbde2 .
 The above commented code is actually part of  this comment already in place.
 
 While it might better to use the commit to add the comments from docs with only addition for the Maistra DEFAULT_NON_FIPS_CIPHER_SUITES.

Actions:
1. Add documentation change from 06f63510c4b5eff7b61d703674fec26b3b1bbde2
2. Un-comment lines in the ssl_socket_test.cc file.
3. Reference the up stream commit hash in the commit message for this PR.

jforce:
@twghu - Should we change the target branch to be 2.3 too?

twghu:
@james-cccc re 2.3 yes we should, I'll add a note to the 2.3 merge issue.

twghu:
Closing this PR as code is stale.  Will add new focused PR.

twghu:
See https://github.com/maistra/envoy/pull/218

PR #181, opened 2022-06-22
[maistra-2.2] MAISTRA-2648: WASM fix for s390x

knm3000:
Fixes WASM on s390x https://issues.redhat.com/browse/MAISTRA-2648
The permanent fixes are:
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/198
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/282
This is the 2nd attempt to enable wasm support for s390x (the 1st one included proxy-wasm PR 198 only and some maistra/proxy tests failed). Now all maistra/proxy tests pass after adding more changes (proxy-wasm PR 282).

Signed-off-by: Konstantin Maksimov konstantin.maksimov@ibm.com

### DIFF
--- a/bazel/external/proxy-wasm-cpp-host-s390x-support.patch
+++ b/bazel/external/proxy-wasm-cpp-host-s390x-support.patch
@@ -1,0 +1,201 @@
+diff --git a/include/proxy-wasm/exports.h b/include/proxy-wasm/exports.h
+index ded6419a47..2451d51530 100644
+--- a/include/proxy-wasm/exports.h
++++ b/include/proxy-wasm/exports.h
+@@ -44,12 +44,13 @@ template <typename Pairs> size_t pairsSize(const Pairs &result) {
+ 
+ template <typename Pairs> void marshalPairs(const Pairs &result, char *buffer) {
+   char *b = buffer;
+-  *reinterpret_cast<uint32_t *>(b) = result.size();
++  bool reverse = "null" != contextOrEffectiveContext()->wasmVm()->runtime();
++  *reinterpret_cast<uint32_t *>(b) = reverse ? htowasm(result.size()) : result.size();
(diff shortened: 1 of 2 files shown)


oschaaf:
It looks like we depend on our forked https://github.com/maistra/proxy-wasm-cpp-host via https://github.com/maistra/envoy/blob/c8f61e7cc630724cbdf82f5c38650090f00cbb5e/bazel/repository_locations.bzl#L1007 and will be patching that with this change.

Imho it would then be cleaner to create a PR to https://github.com/maistra/proxy-wasm-cpp-host with these changes and then update the dependency version here once those land.. wdyt?

knm3000:
Agree, for maistra-2.2 it makes sense move these changes to https://github.com/maistra/proxy-wasm-cpp-host, will do that.
Information for future maistra releases:
It looks like we switched to using maistra/proxy-wasm-cpp-host instead of proxy-wasm/proxy-wasm-cpp-host for openssl reason only. However, proxy-wasm/proxy-wasm-cpp-host starting from Jan 2022 version supports any ssl implementation installed on host (in addition to boringssl, which is still used by default):
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/219
This is the example how to setup openssl for a particular platform:
https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/220
Maybe in maistra-2.3 and later we could use that.
maistra-2.2 has proxy-wasm-cpp-host version from Aug 2021, so it doesn't include those commits, and therefore maistra/proxy-wasm-cpp-host is still needed for maistra-2.2.

oschaaf:
Thanks! 
/cc @dcillera who is working on 2.3

knm3000:
Opened a new PR https://github.com/maistra/proxy-wasm-cpp-host/pull/1 instead of this one

PR #187, opened 2022-07-13
Make RELEASE build by default

knm3000:
Building and testing with optimization potentially could help to detect hidden problems. Other repos (envoyproxy/envoy and maistra/proxy) run RELEASE build by default.

### DIFF
--- a/maistra/bazelrc
+++ b/maistra/bazelrc
@@ -5,6 +5,7 @@ build --cxxopt -w
 
 build --copt -DOPENSSL_IS_BORINGSSL=0
 build --verbose_failures
+build -c opt
 
 # As of today we do not support QUIC/HTTP3 -- hence we exclude it from the build, always.
 build --deleted_packages=test/common/quic,test/common/quic/platform


maistra/bazelrc
```diff
@@ -5,6 +5,7 @@ build --cxxopt -w
 
 build --copt -DOPENSSL_IS_BORINGSSL=0
 build --verbose_failures
+build -c opt
```
oschaaf, line 8: I feel `build:ci-config -c opt` would be better; that would make this the default for our ci-environment.. and maybe a one liner explaining why we do this would be helpful too for future maintenance.

oschaaf (changes requested):
To be honest, I'm not sure I am comfortable about the *testing* with `-c opt` part, as that might cause a bunch of debug assertions to not fire when executing tests? 

I think we ought to build and test both, but given that as of today we only test one (due to CI resource constraints probably) .. I am more comfortable with the status quo.. We could build with -c opt and leave the tests as they are as a middle ground? (though this would impact CI runtime duration a lot, I think, as the CI test step no longer can benefit from the build step that was done earlier).

Having said that, I may not have all the context.. could you elaborate for a bit on the kind of hidden problems building/testing with -c opt would surface?

jwendell (commented):
I don't have a strong opinion. I think Otto's suggestion might work for all scenarios. We can always override those flags by putting them in a local `user.bazelrc` or `~/.bazelrc` if necessary.

In proxy, we are already building using `-c opt`. See https://github.com/maistra/proxy/blob/maistra-2.2/maistra/ci/common.sh#L18 and https://github.com/maistra/proxy/blob/maistra-2.2/.bazelrc#L34

maistra/bazelrc
twghu: I agree with Otto on this,  while a RELEASE configuration is ideal we need to identify the problem this change addresses.

knm3000: There was a proxy crash problem https://issues.redhat.com/browse/OSSM-1526 that happens only for the release proxy build. But it is a different scenario: the go tests call envoy proxy binary (which is built in release mode). Here in envoy repo we have only C++ tests, that are built as a separate binary, so that proxy test scenario may not be relevant for envoy repo tests.

knm3000:
Thanks for your opinions, so the disadvantages of using release by default:  
- debug assertions will not trigger
- the usage of limited CI resources will increase (even if build release binary and debug tests)

Since we don't have a concrete problem, that can be solved by switching to release builds, probably we don't need to make any changes now.

maistra/bazelrc
oschaaf: I see; I am blissfully unaware of the details around that :-) but my first (perhaps naive) thought is that often such "heisenbugs" that only occur when certain compiler flags are enabled, are in fact caused by undefined behaviour.. In which case I'd rather see us run Envoy's tests with ubsan in CI (the fact that `-c opt` teases out such issues occasionally is not guaranteed, but ubsan should reliably point out certain types of undefined behaviour as long as test coverage is decent).

This thought doesn't just apply here, thinking about it, it also applies to proxy. In retrospect, I'm actually curious if running proxy tests with sanitizers enabled would have caught https://istio.io/latest/news/security/istio-security-2022-005/ earlier.

PR #189, opened 2022-07-26
Tests fixing/disabling and script cleaning

dcillera:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/config/utility.cc
+++ b/test/config/utility.cc
@@ -950,16 +950,6 @@ void ConfigHelper::setPorts(const std::vector<uint32_t>& ports, bool override_po
   }
   ASSERT(skip_port_usage_validation_ || port_idx == ports.size() || eds_hosts ||
          original_dst_cluster || custom_cluster || bootstrap_.dynamic_resources().has_cds_config());
-  if (!connect_timeout_set_) {
-#ifdef __APPLE__
-    // Set a high default connect timeout. Under heavy load (and in particular in CI), macOS
-    // connections can take inordinately long to complete.
-    setConnectTimeout(std::chrono::seconds(30));
-#else
-    // Set a default connect timeout.
-    setConnectTimeout(std::chrono::seconds(5));
-#endif
-  }
 }
 
 void ConfigHelper::setSourceAddress(const std::string& address_string) {
--- a/test/config/utility.cc
+++ b/test/config/utility.cc
@@ -960,8 +960,6 @@ void ConfigHelper::setPorts(const std::vector<uint32_t>& ports, bool override_po
     setConnectTimeout(std::chrono::seconds(5));
 #endif
   }
-
-  finalized_ = true;
 }
 
 void ConfigHelper::setSourceAddress(const std::string& address_string) {
--- a/test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
+++ b/test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
@@ -227,6 +227,12 @@ INSTANTIATE_TEST_SUITE_P(IpVersions, ProxyFilterIntegrationTest,
 // A basic test where we pause a request to lookup localhost, and then do another request which
 // should hit the TLS cache.
 TEST_P(ProxyFilterIntegrationTest, RequestWithBody) {
+  int64_t original_usec = dispatcher_->timeSource().monotonicTime().time_since_epoch().count();
+
+  config_helper_.prependFilter(fmt::format(R"EOF(
+  name: stream-info-to-headers-filter
+)EOF"));
+
   initializeWithArgs();
   codec_client_ = makeHttpConnection(lookupPort("http"));
   const Http::TestRequestHeaderMapImpl request_headers{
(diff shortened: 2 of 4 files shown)


test/config/utility.cc
```diff
@@ -960,8 +960,6 @@ void ConfigHelper::setPorts(const std::vector<uint32_t>& ports, bool override_po
     setConnectTimeout(std::chrono::seconds(5));
 #endif
   }
-
-  finalized_ = true;
```
oschaaf, line 964: Can you explain why we need this?

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
```diff
@@ -264,7 +279,8 @@ TEST_P(ProxyFilterIntegrationTest, RequestWithUnknownDomain) {
   auto response = codec_client_->makeHeaderOnlyRequest(default_request_headers_);
   ASSERT_TRUE(response->waitForEndStream());
   EXPECT_EQ("503", response->headers().getStatusValue());
-  EXPECT_THAT(waitForAccessLog(access_log_name_), HasSubstr("dns_resolution_failure"));
```
oschaaf, line 267: accidental log line to stdout?

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
```diff
@@ -227,6 +227,12 @@ INSTANTIATE_TEST_SUITE_P(IpVersions, ProxyFilterIntegrationTest,
 // A basic test where we pause a request to lookup localhost, and then do another request which
 // should hit the TLS cache.
 TEST_P(ProxyFilterIntegrationTest, RequestWithBody) {
+  int64_t original_usec = dispatcher_->timeSource().monotonicTime().time_since_epoch().count();
```
oschaaf, line 230: can you please add a code-level comment, per change required to fix a test, to document why we diverge from vanilla Envoy 1.20 (what we did to resolve the test / and maybe an analysis of why we think it's needed)?

some day, someone is going to read these changes again (perhaps when attempting to migrate to the next Envoy version), and then such comments will be super helpful.

oschaaf (changes requested):
great stuff! flushing out some comments

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
dcillera: Yes, it has been removed in the next commit

test/config/utility.cc
dcillera: This is as in Maistra 2.2

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
dcillera: Yes, it has been removed in the next commit

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
dcillera: In [f98723c] I refined the modification and added a comment

test/config/utility.cc
oschaaf: When I compare this full file (`test/config/utility.cc`) with what's in vanilla upstream [1] I get a bit confused, I observe more changes then I would intuitively expect.. can you double check if that is to be expected or not? 

So having said that, vanilla Envoy 1.20 *does* have this line of code, I wonder why we dropped it in our code base. So something that would be interesting to know is if we can figure out if and why we need to diverge (maybe git blame can help here?)

[1] https://raw.githubusercontent.com/envoyproxy/envoy/release/v1.20/test/config/utility.cc

dcillera: There are many changes indeed, but I have just merged all the stuff from 1.22. It's hard to say if all of them are really needed but, as far as I have understood, all changes from upstream have to be merged unless they clash with Maistra specific features (this relates mostly with OpenSSL).
As regards the modification in this specific commit, I have checked:
- In 1.20 and 2.2 that line of code (finalized_ = true;) was present, together with  timer starting;
- In 1.22 that line was removed, together with the timer.
I think in 1.22 they have rearranged the code in a way that those lines must be removed. In fact, if they are kept, the multi_envoy test fails.
 Then, I think we should align our code to 1.22 and remove also the timer.

twghu: I have checked the code changes and differences from v1.22/2.2 through v1.22/2.3 and also 2.2->2.3 for this file.  
All differences except one are accounted for.

The outstanding change in this file in the 2.3 branch is the the finalization code was moved by Alyssa from the setPorts() to finalize().  Consequently the lines of code in setPorts() should be removed.

```
diff test/config/utility.cc ~/work/envoyproxy/envoy/test/config/utility.cc
953,964d952
<   if (!connect_timeout_set_) {
< #ifdef __APPLE__
<     // Set a high default connect timeout. Under heavy load (and in particular in CI), macOS
<     // connections can take inordinately long to complete.
<     setConnectTimeout(std::chrono::seconds(30));
< #else
<     // Set a default connect timeout.
<     setConnectTimeout(std::chrono::seconds(5));
< #endif
<   }
< 
<   finalized_ = true;
```
    
This code was moved to finalize().

jwendell (approved):

twghu (approved):

oschaaf (approved):

PR #191, opened 2022-07-27
Added big-endian support into TestByteOrder

knm3000:
https://github.com/envoyproxy/envoy/pull/22353

### DIFF
--- a/test/common/network/utility_test.cc
+++ b/test/common/network/utility_test.cc
@@ -530,22 +530,29 @@ TEST(PortRangeListTest, Normal) {
   }
 }
 
-// TODO(ccaraman): Support big-endian. These tests operate under the assumption that the machine
-// byte order is little-endian.
 TEST(AbslUint128, TestByteOrder) {
+#if defined(ABSL_IS_BIG_ENDIAN)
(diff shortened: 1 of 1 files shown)


oschaaf:
This LGTM, but I wonder about the following: This is the equivalent of upstream commit https://github.com/envoyproxy/envoy/commit/e834c24e061b710348ffd72016d5d1069698b4ff, right? I suspect that using `cherry-pick` to include that might make the next rebase a tiny fraction easier, and might therefore be preferred.. any thoughts on that? @twghu and/or @dcillera .. any thoughts/preference?

oschaaf (approved):

knm3000:
Correct, this is the equivalent of upstream commit https://github.com/envoyproxy/envoy/commit/e834c24e061b710348ffd72016d5d1069698b4ff
Let me know if I should remove this maistra commit and make `cherry-pick` instead.

jwendell:
In upstream, backport it to 1.20 and 1.22 branches. Then on the next sync we will get it for free

knm3000:
There is no urgency to backport and include s390x fix for this particular envoy test into 2.2.

PR #179, opened 2022-06-18
Add scripts to support test execution with sanitizers enabled.

oschaaf:
- Refactors run-ci.sh a little bit to make it easier to use in dev flows
- Add new scripts to support executing with asan/tsan/msan enabled

Signed-off-by: Otto van der Schaaf <ovanders@redhat.com>

PR #194, opened 2022-08-04
OSSM-1830 Sync with upstream 1.22

jwendell:
And add minor fixes (see last commits)


maistra/run-ci.sh
```diff
@@ -4,34 +4,12 @@ set -e
...
-fi
-export ARCH
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+source "$DIR/common.sh"
 
 export BUILD_SCM_REVISION="Maistra PR #${PULL_NUMBER:-undefined}"
 export BUILD_SCM_STATUS="SHA=${PULL_PULL_SHA:-undefined}"
 
-COMMON_FLAGS="\
-    --config=clang \
```
oschaaf, line 19: I wouldn't be surprised if we might actually need to have `config` set to `clang` for 2.3, but let's see what CI says about that:

```diff
diff --git a/maistra/bazelrc b/maistra/bazelrc
index bfa3e5ada2..22ffa1bb69 100644
--- a/maistra/bazelrc
+++ b/maistra/bazelrc
@@ -6,6 +6,8 @@ build --cxxopt -w
 build --copt -DOPENSSL_IS_BORINGSSL=0
 build --verbose_failures
 
+build --config=clang
+
 # As of today we do not support QUIC/HTTP3 -- hence we exclude it from the build, always.
 build --deleted_packages=test/common/quic,test/common/quic/platform
 build "--//bazel:http3"=false
```

@dcillera do you remember if you ended up needing this?

oschaaf (approved):

maistra/run-ci.sh
dcillera: No, I don't remember that was needed, but I have used that line to locally and temporarily enable ASAN in a build:
 build --config=clang-asan
In any case, I know you have provided this feature in #179.

jwendell: JFYI, I added this back, we got an error on previous CI build that I think was because of this.

oschaaf: Yes -- I thought we'd run into some build error earlier when we included #179 for 2.3 (specifically the `tcmalloc` dependency) ... and that adding back `--config=clang` resolved it. Checking CI I expect we will pass now with Jonh his latest change.

PR #196, opened 2022-08-30
Add bazel test Throttle resources for CI

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/maistra/bazelrc
+++ b/maistra/bazelrc
@@ -20,6 +20,9 @@ build:aarch64 --linkopt=-fuse-ld=lld
 build:ci-config --local_ram_resources=12288
 build:ci-config --local_cpu_resources=6
 build:ci-config --jobs=3
+test:ci-config --local_ram_resources=12288
+test:ci-config --local_cpu_resources=6
+test:ci-config --jobs=3
 
 # Colored ouput messes with some of the logging systems in CI, so we disable that.
 build:ci-config --color=no


yxun:
When I run maistra-2.2 maistra/envoy maistra/run-ci.sh in an OCP pod, I set the env variable CI="true" in the container.
At the beginning of the log , I can see --config=ci-config

```
++ '[' -n true ']'
++ COMMON_FLAGS+=' --config=ci-config '
++ '[' -n '' ']'
++ '[' -n '' ']'
```

However, in the middle of the run-ci.sh exection, my pod CPU jumped and reached request limits.Then OCP killed an integration test execution when it requested more CPU than the limit.

image

From the reference, it looks like we only set the resource throttle in the `bazel build ...` part. But we haven't set any resource throttle in the `bazel test ...` part yet.
https://bazel.build/reference/command-line-reference

oschaaf:
We discussed this on Slack earlier: I feel that tuning these resource limits should be done only when our CI actually needs it. Having said that, maybe a bazel comment indicating that said config line only throttles build and not test execution is worth it?

I'm curious if the CI failure is a flake of some sorts, I have a hard time associating them to this change.. so:

/retest

maistra/bazelrc
```diff
@@ -20,6 +20,9 @@ build:aarch64 --linkopt=-fuse-ld=lld
 build:ci-config --local_ram_resources=12288
 build:ci-config --local_cpu_resources=6
 build:ci-config --jobs=3
+test:ci-config --local_ram_resources=12288
```
jwendell, line 23: This should not be necessary as `test` inherits is configuration from `build`: https://bazel.build/run/bazelrc#option-defaults

oschaaf:
Closing this: with earlier testing, me and @jwendell  have observed that the change proposed here actually is a no-op (as suggested in a comment above).

yxun:
https://bazel.build/docs/user-manual#running-tests
"By default, this command performs simultaneous build and test activity"
That might explain why the cpu usage is over the config value 6.
I will try some workaround thanks.

PR #200, opened 2022-09-07
Split build tests and run tests steps in run-ci

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

Commit Message:
Split build tests and run tests steps in CI

Additional Description:
This change is useful for reducing the resources usage when compiling and running tests.
I tested this change can avoid CPU usage spike in my OCP environment and help control the resources in the limited request level.

Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/maistra/run-ci.sh
+++ b/maistra/run-ci.sh
@@ -14,6 +14,16 @@ time bazel build \
 echo "Build succeeded. Binary generated:"
 bazel-bin/source/exe/envoy-static --version
 
+# By default, `bazel test` command performs simultaneous
+# build and test activity.
+# The following build step helps reduce resources usage
+# by compiling tests first.
+# Build tests
+time bazel build \
+  ${COMMON_FLAGS} \
+  --build_tests_only \
+  //test/...
+
 # Run tests
 time bazel test \
   ${COMMON_FLAGS} \


maistra/run-ci.sh
```diff
@@ -14,6 +14,11 @@ time bazel build \
 echo "Build succeeded. Binary generated:"
 bazel-bin/source/exe/envoy-static --version
 
+# Build tests
+time bazel build \
```
oschaaf, line 22: Can we add a doc comment to explain why we do split the build and test steps for future readers?

maistra/run-ci.sh
```diff
@@ -14,6 +14,11 @@ time bazel build \
...
+  ${COMMON_FLAGS} \
+  //test/...
```
jwendell, line 25: Perhaps add `--build_tests_only` too?

jwendell (commented):
Was this of any help at all?

maistra/run-ci.sh
yxun: I tried this change with local disk cache. It helps avoid a CPU usage spike. 
hm, I assume all files under the //test/ are tests. I can add --build_tests_only in that line

oschaaf (approved):

PR #204, opened 2022-09-21
Small tweaks to build system

jwendell:
- Remove duplicated flag `--config=clang` that causes a WARNING
- Move the ipv4 flag for tests to be applied everywhere, not only in CI
- Build tests before running them. This might help with CPU usages


oschaaf (approved):
Great, LGTM

PR #207, opened 2022-09-30
MAISTRA-2756: disable tests that fail on s390x

knm3000:


knm3000:
Currently we don't have envoy CI for s390x, no benefit from this change.

PR #211, opened 2022-10-17
OSSM-1801 Fixed pathlen constraint on test certificate and re enabled test

tedjpoole:
There are four test certificates under `test/config/integration/certs` that form a chain which fails to verify correctly because of a [path length constraint](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.9) in one of the intermediate certificates. The certificate chain is as follows:

	cacert.pem
		> intermediate_cacert.pem [ pathlen: 0 ]
			> intermediate_ca_2cert.pem [ pathlen: 0 ]
				> client2cert.pem

The `pathlen : 0`, in `intermediate_cacert.pem` indicates that it must not be followed by any non-self-issued intermediate CA certificates. Therefore, the 'intermediate_ca_2cert.pem' certificate violates that constraint, leading to the test failure. This PR updates the pathlen constraint in `intermediate_cacert.pem` from `0` to `1`, so that the chain verifies ok.


jwendell:
So, if I understand it correctly, this is an issue present in upstream too, right? I'm curious why it fails only in our fork.

Anyway, I'd love to see this being submitted upstream.

jwendell (approved):

PR #220, opened 2022-12-16
OSSM-2208: SSL handshake may fail with invalid OCSP response

dcillera:
Signed-off-by: Dario dcillera@redhat.com
Commit Message: OSSM-2208 fixing
Additional Description: Enable TestFilterMultipleCertsFilterByOcspPolicyFallbackOnFirst test and add  a cipher-suite on server side configuration as done in the upstream.
Risk Level: Low
Testing: Done


jwendell:
Please use proper commit messages, and remove the clutter in the description

twghu:
While this change causes the test to pass, I'm not convinced it's proving what we think it is.  
Refer to comments in Issue 1667 for further actions we need to take.

dcillera:
Closed because the  related commit was not a valid solution.

PR #230, opened 2023-02-16
OSSM-3429: Fix a compiler error

jwendell:
Recently CI saw errors like this:
```
bazel-out/k8-fastbuild/bin/external/org_llvm_releases_compiler_rt/_virtual_includes/fuzzed_data_provider/fuzzer/FuzzedDataProvider.h:194:38: error: no member named 'numeric_limits' in namespace 'std'
  return ConsumeIntegralInRange(std::numeric_limits<T>::min(),
                                ~~~~~^
```

Apparently this is due to a incompatibility between the `compiler-rt` dependency and the clang version we use (14.0).

This is fixed by updating `compiler-rt` to match clang 14. Actually This commit cherry-picks the relevant changes done in upstream a while ago: https://github.com/envoyproxy/envoy/pull/21135


jwendell:
I'm still not sure why this became an issue only now and not before.

Another approach for this issue (and future issues) would stick with clang 12 which is the version upstream 1.22 uses. But this is not ideal since the product is built using clang 14. Hence we also use clang 14 in community CI, to catch errors like this before we attempt to perform a product build (i.e. too late).

twghu (approved):

jwendell:
Incorporated in https://github.com/maistra/envoy/pull/234

PR #231, opened 2023-02-20
Use local python instead of bazel toolchain python

knm3000:
After commit https://github.com/envoyproxy/envoy/pull/20432 python interpreter from rules_python bazel toolchain is used instead of the local python. The problem is that there is no s390x support (as well as p) in rules_python repos: https://github.com/bazelbuild/rules_python/blob/main/python/versions.bzl
https://github.com/indygreg/python-build-standalone/releases
and therefore bazel build fails on s390x `'@python3_10//': No platform declared for host OS linux on arch s390x`.


twghu (approved):

twghu:
In retrospect, should this type of change be wrapped in a "is this a s390 type build" so that it has no effect on builds that don't have this error?

PR #256, opened 2023-04-11
[maistra-2.3] OSSM-2208: With multiple cert chains, SSL handshake may fail

dcillera:
Fixed OSSM-2208: With multiple cert chains, SSL handshake may fail if one with invalid OCSP response is selected

### DIFF
--- a/source/extensions/transport_sockets/tls/context_impl.cc
+++ b/source/extensions/transport_sockets/tls/context_impl.cc
@@ -981,19 +981,27 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
 
 int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
   const bool client_ocsp_capable = isClientOcspCapable(ssl);
-  // returns server cert selected for this connection
-  // see https://www.openssl.org/docs/man1.1.1/man3/SSL_set_tlsext_status_ocsp_resp.html
-  // for details
-  auto* selected_cert = SSL_get_certificate(ssl);
-  const auto& cert_context = certificateContext(selected_cert);
-  auto ocsp_staple_action = ocspStapleAction(cert_context, client_ocsp_capable);
+  
+  // Loop on all certificates to find at least a good one
+  const CertContext* selected_cert_context = nullptr;
+  auto  ocsp_staple_action = OcspStapleAction::Fail;
+  for(const auto& cert_context : tls_context_.cert_contexts_) {
+    RELEASE_ASSERT(SSL_select_current_cert(ssl,cert_context.cert_chain_.get()),
+		  "SSL_select_current_cert() failure");
+    ocsp_staple_action = ocspStapleAction(cert_context, client_ocsp_capable);
+    if (ocsp_staple_action == OcspStapleAction::Fail) {
+      continue;
+    }
+    selected_cert_context = &cert_context;
+    break;
+  }
 
   switch (ocsp_staple_action) {
   case OcspStapleAction::Staple: {
     // We avoid setting the OCSP response if the client didn't request it, but doing so is safe.
-    RELEASE_ASSERT(cert_context.ocsp_response_,
+    RELEASE_ASSERT(selected_cert_context->ocsp_response_,
                    "OCSP response must be present under OcspStapleAction::Staple");
-    const std::vector<uint8_t>& raw_bytes = cert_context.ocsp_response_->rawBytes();
+    const std::vector<uint8_t>& raw_bytes = selected_cert_context->ocsp_response_->rawBytes();
     const std::size_t raw_bytes_size = raw_bytes.size();
     unsigned char* raw_bytes_copy = static_cast<unsigned char *>(OPENSSL_memdup(raw_bytes.data(), raw_bytes_size));
     if (raw_bytes_copy == nullptr) { 
@@ -1015,7 +1023,6 @@ int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
   case OcspStapleAction::ClientNotCapable:
     return SSL_TLSEXT_ERR_NOACK;
   }
-
   return SSL_TLSEXT_ERR_OK;
 }
 
--- a/test/extensions/transport_sockets/tls/ssl_socket_test.cc
+++ b/test/extensions/transport_sockets/tls/ssl_socket_test.cc
@@ -6262,15 +6262,7 @@ TEST_P(SslSocketTest, TestConnectionFailsWhenCertIsMustStapleAndResponseExpired)
   testUtil(test_options.setExpectedServerStats("ssl.ocsp_staple_failed").enableOcspStapling());
 }
 
-// TODO (dmitri-d) we currently rely on OpenSSL to setup server-side certificate chain to use.
-// OpenSSL selection process doesn't take into account presence and status of an OCSP response.
-// This test failure under OpenSSL as only one of the configured certificate chains has a valid
-// OSCP response, which is expected to be used.
-// A possible approach is to use a cert_cb (see
-// https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_cert_cb.html) and check OCSP response
-// validity then. Using this callback raises a question of cert chain compatibility with the client
-// side and how to handle it.
-TEST_P(SslSocketTest, DISABLED_TestFilterMultipleCertsFilterByOcspPolicyFallbackOnFirst) {
+TEST_P(SslSocketTest, TestFilterMultipleCertsFilterByOcspPolicyFallbackOnFirst) {
   const std::string server_ctx_yaml = R"EOF(
   common_tls_context:
     tls_certificates:


source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -4,7 +4,6 @@
 #include <memory>
 #include <string>
 #include <vector>
-
```
tedjpoole: Unnecessary deletion

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -751,6 +750,7 @@ ServerContextImpl::ServerContextImpl(Stats::Scope& scope,
     }
   }
 
+
```
tedjpoole: Unnecessary addition

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -979,21 +979,32 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
   PANIC_DUE_TO_CORRUPT_ENUM;
 }
 
+
```
tedjpoole: Unnecessary addition

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -979,21 +979,32 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
...
 int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
   const bool client_ocsp_capable = isClientOcspCapable(ssl);
-  // returns server cert selected for this connection
-  // see https://www.openssl.org/docs/man1.1.1/man3/SSL_set_tlsext_status_ocsp_resp.html
-  // for details
-  auto* selected_cert = SSL_get_certificate(ssl);
-  const auto& cert_context = certificateContext(selected_cert);
-  auto ocsp_staple_action = ocspStapleAction(cert_context, client_ocsp_capable);
+
```
tedjpoole, line 984: Are you sure that iterating with `SSL_CTX_set_current_cert(ctx, SSL_CERT_SET_FIRST/NEXT)` is going to match the iteration over `tls_context_.cert_contexts_`? What if the two collections have different members, or they are ordered differently?

I think it would be safer to use just one method of iteration i.e. either:

(1) Iterate with `SSL_set_current_cert(ctx, SSL_CERT_SET_FIRST/NEXT)` and use `SSL_get0_chain_certs()` to get the current certificate to be checked for suitability.

OR

(2) Iterate over `tls_context_.cert_contexts_` to find a suitable certificate chain, and then use something like `SSL_set0_chain()` to set it as the current one.

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -979,21 +979,32 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
...
+      selected_cert_context = &cert_context;
+      ocsp_staple_action = action;
+      break;
+  }
 
   switch (ocsp_staple_action) {
   case OcspStapleAction::Staple: {
     // We avoid setting the OCSP response if the client didn't request it, but doing so is safe.
-    RELEASE_ASSERT(cert_context.ocsp_response_,
+    RELEASE_ASSERT(selected_cert_context->ocsp_response_,
```
tedjpoole, line 1002: This is setting the OCSP reponse from a certificate that may or may not match the most recent one selected by `SSL_CTX_set_current_cert()`.

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -979,21 +979,32 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
...
-  // see https://www.openssl.org/docs/man1.1.1/man3/SSL_set_tlsext_status_ocsp_resp.html
-  // for details
-  auto* selected_cert = SSL_get_certificate(ssl);
-  const auto& cert_context = certificateContext(selected_cert);
-  auto ocsp_staple_action = ocspStapleAction(cert_context, client_ocsp_capable);
+  
+  // Fix OSSM-2208 by looping on all certificates to find at least a good one
+  const CertContext* selected_cert_context = &tls_context_.cert_contexts_[0];
+  auto  ocsp_staple_action = ocspStapleAction(tls_context_.cert_contexts_[0], client_ocsp_capable);   
+  SSL_CTX_set_current_cert(tls_context_.ssl_ctx_.get(), SSL_CERT_SET_FIRST);
```
tedjpoole: Check return value

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -979,21 +979,32 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
...
+  for(const auto& cert_context : tls_context_.cert_contexts_) {
+      auto action = ocspStapleAction(cert_context, client_ocsp_capable);
+      
+      if (action == OcspStapleAction::Fail) {
+          SSL_CTX_set_current_cert(tls_context_.ssl_ctx_.get(), SSL_CERT_SET_NEXT);
```
tedjpoole: Check return value

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -1015,10 +1026,10 @@ int ServerContextImpl::handleOcspStapling(SSL* ssl, void*) {
   case OcspStapleAction::ClientNotCapable:
     return SSL_TLSEXT_ERR_NOACK;
   }
-
   return SSL_TLSEXT_ERR_OK;
 }
 
+
```
tedjpoole: Unnecessary addition

source/extensions/transport_sockets/tls/context_impl.cc
dcillera: That's mandatory!

source/extensions/transport_sockets/tls/context_impl.cc
dcillera: Your objection makes sense. Digging into the code, I concluded that the two collections have the same members (loaded using `SSL_CTX_use_certificate`) but they may be ordered differently. 
Now trying your suggested solution (2).

dcillera: Implementing a solution that calls: SSL_select_current_cert instead of SSL_set_current_cert

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -981,19 +981,29 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
...
-  // for details
-  auto* selected_cert = SSL_get_certificate(ssl);
-  const auto& cert_context = certificateContext(selected_cert);
-  auto ocsp_staple_action = ocspStapleAction(cert_context, client_ocsp_capable);
+  
+  // Fix OSSM-2208 by looping on all certificates to find at least a good one
+  const CertContext* selected_cert_context = &tls_context_.cert_contexts_[0];
+  auto  ocsp_staple_action = ocspStapleAction(tls_context_.cert_contexts_[0], client_ocsp_capable);
+  for(const auto& cert_context : tls_context_.cert_contexts_) {
+      if(!SSL_select_current_cert(ssl,tls_context_.cert_contexts_[0].cert_chain_.get())) {
```
tedjpoole: This line is just repeatedly setting the current certificate chain to be the same one from `tls_context_.cert_contexts_[0]` every time, regardless of the value of `cert_context` from the for loop.

I think you want to replace it with: `SSL_select_current_cert(ssl, cert_context.cert_chain_.get())` otherwise you're just selecting the same (first) certificate every time round the loop, regardless of the `cert_context` loop variable.

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -981,19 +981,29 @@ OcspStapleAction ServerContextImpl::ocspStapleAction(const CertContext& cert_con
...
   const bool client_ocsp_capable = isClientOcspCapable(ssl);
-  // returns server cert selected for this connection
-  // see https://www.openssl.org/docs/man1.1.1/man3/SSL_set_tlsext_status_ocsp_resp.html
-  // for details
-  auto* selected_cert = SSL_get_certificate(ssl);
-  const auto& cert_context = certificateContext(selected_cert);
-  auto ocsp_staple_action = ocspStapleAction(cert_context, client_ocsp_capable);
+  
+  // Fix OSSM-2208 by looping on all certificates to find at least a good one
+  const CertContext* selected_cert_context = &tls_context_.cert_contexts_[0];
```
tedjpoole: The `selected_cert_context` and `ocsp_staple_action` variables don't really need to be initialised before the loop. They will get initialised to those same values in the first iteration of the loop anyway.

dcillera: Ok, but I had taken that code from the upstream Envoy 1.24 (function selectTlsContext).

dcillera: /retest

tedjpoole:
I've just made some minor changes to indentation, and removed Dmitri's comment from the test because it is no longer relevant.

tedjpoole:
LGTM. I tried swapping the order of the server certificates in the test, and verified that it does then go round your loop twice instead of once, to find the appropriate certificate chain to use.

tedjpoole (approved):

PR #264, opened 2023-04-17
[Maistra-2.2] OSSM-2208: With multiple cert chains, SSL handshake may fail

dcillera:
OSSM-2208: With multiple cert chains, SSL handshake may fail if one with invalid OCSP response is selected (#256)


tedjpoole (approved):
LGTM

PR #271, opened 2023-05-11
Added support for setting Bazel's --experimental_remote_downloader option

tedjpoole:
This is beneficial when using the [buchgr/bazel-remote](https://github.com/buchgr/bazel-remote) cache, to reduce the number of external downloads over the internet, and therefore speed up builds.

### DIFF
--- a/maistra/common.sh
+++ b/maistra/common.sh
@@ -34,3 +34,7 @@ if [ -n "${BAZEL_REMOTE_CACHE}" ]; then
 elif [ -n "${BAZEL_DISK_CACHE}" ]; then
   COMMON_FLAGS+=" --disk_cache=${BAZEL_DISK_CACHE} "
 fi
+
+if [ -n "${BAZEL_EXPERIMENTAL_REMOTE_DOWNLOADER}" ]; then
+  COMMON_FLAGS+=" --experimental_remote_downloader=${BAZEL_EXPERIMENTAL_REMOTE_DOWNLOADER}"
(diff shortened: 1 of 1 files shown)


maistra/common.sh
```diff
@@ -33,4 +33,8 @@ if [ -n "${BAZEL_REMOTE_CACHE}" ]; then
   COMMON_FLAGS+=" --remote_cache=${BAZEL_REMOTE_CACHE} "
 elif [ -n "${BAZEL_DISK_CACHE}" ]; then
   COMMON_FLAGS+=" --disk_cache=${BAZEL_DISK_CACHE} "
-fi
\ No newline at end of file
+fi
+
+if [ -n "${BAZEL_EXPERIMENTAL_REMOTE_DOWNLOADER}" ]; then
```
jwendell, line 38: Hey @tedjpoole out of curiosity, what's the difference between this flag and `--remote_cache` as specified here: https://github.com/buchgr/bazel-remote#configuring-bazel ?

tedjpoole: `--experimental_remote_downloader` applies to downloads, whereas the `--remote_cache` option applies to build actions. For me, with a relatively slow internet connection, this speeds up downloading all the third party packages.

jwendell: Ah, of course. It's clear now, thanks! So it's essentially a proxy cache.

jwendell (approved):

PR #272, opened 2023-05-16
Client Cipher update for OSSM-1667

twghu:
Updates to align Maistra OpenSSL with Upstream.


tedjpoole:
I thought we were going to add a feature flag to re-enable the old ciphers?

twghu:
That would be an ideal option, but at this time will not implemented.

tedjpoole (approved):

PR #275, opened 2023-05-31
Bump to the version used by Istio 1.16.5

jwendell:


jwendell:
Changes from upstream that caused conflict were:
```diff
- deps = ["@boringssl//:ssl"]
+ deps = ["//bazel:boringssl"]
```

But since we did 
```diff
- deps = ["@boringssl//:ssl"]
+ deps = ["@openssl//:openssl-lib"]
```

As part of our downstream changes, we are safe to ignore such changes. We are not using boringssl libs anyway.

dcillera (approved):

PR #295, opened 2023-10-13
[Maistra 2.3] CVE - Denial of service when using HTTP/2 protocol  

dcillera:
Cherry picked from the Upstream the commits related to the CVE "Denial of service when using HTTP/2 protocol" and  other commits necessary to a correct build & test.


source/common/runtime/runtime_features.cc
```diff
@@ -80,6 +80,8 @@ RUNTIME_GUARD(envoy_reloadable_features_validate_connect);
 RUNTIME_GUARD(envoy_reloadable_features_validate_upstream_headers);
 RUNTIME_GUARD(envoy_restart_features_explicit_wildcard_resource);
 RUNTIME_GUARD(envoy_restart_features_no_runtime_singleton);
+RUNTIME_GUARD(envoy_restart_features_remove_runtime_singleton);
```
jwendell, line 83: Is this correct? The line above seems to clash with the previous one? Not sure how harm this is though...

changelogs/current.yaml
```diff
@@ -1,9 +1,22 @@
 date: Pending
 
 behavior_changes:
+<<<<<<< HEAD
```
jwendell, line 4: this and === and >>> lines below seem to be a merge conflict leftover. Harmless in this pr as it's just doc. Can be fixed later.

changelogs/current.yaml
```diff
@@ -16,7 +16,17 @@ behavior_changes:
...
 >>>>>>> 01b62a78e9 (Close HTTP connections that prematurely reset streams)
+=======
+- area: http
+  change: |
+    Add runtime flag ``http.max_requests_per_io_cycle`` for setting the limit on the number of HTTP requests processed
+    from a single connection in a single I/O cycle. Requests over this limit are processed in subsequent I/O cycles. This
+    mitigates CPU starvation by connections that simultaneously send high number of requests by allowing requests from other
+    connections to make progress. This runtime value can be set to 1 in the presence of abusive HTTP/2 or HTTP/3 connections.
+    By default this limit is disabled.
+>>>>>>> b8f05c054c (Limit on the number of HTTP requests processed from a connection in an I/O cycle)
```
jwendell, line 29: leftovers from merge conflict, can be fixed later

jwendell (approved):
just one minor comment, not sure how important it is.

approving to get it expedite, we can fix any eventual mistakes in a later pr

source/common/runtime/runtime_features.cc
dcillera: Probably not correct, but it should be harmless as only the no_runtime_singleton is used in the code.

PR #304, opened 2024-01-30
Initial work for 2.5

jwendell:
This is based on Dario's branch, plus one commit with a few adjustments to CI scripts and removal of some leftover files from rebase.


jwendell:
- These envoy prechecks will be go away after this PR is merged, I think. Last commit in this PR removes everything related to them.
- Our CI script only builds Envoy for now. Tests are not running. We should re-enable them in a follow-up PR, when we will fix the failures.

jwendell:
@yxun the tests passed but the job failed at some post action. Can you please take a look?

yxun:
> @yxun the tests passed but the job failed at some post action. Can you please take a look?

What are that set of Github CI jobs ? It looks like in the ci directory. They are triggered by Github or bot

yxun:
"Process did not finish before 10m0s timeout"  we have some timeout limit in cleanup step. The namespace deletion was stuck. Actually, we don't need that step. The deprovision cluster step is enough for cleanup.

yxun:
@jwendell  https://github.com/openshift/release/pull/48228

twghu (approved):
Ok as initial PR.

PR #306, opened 2024-02-14
[maistra-2.4] OSSM-5850: ext_authz: ensure both raw body/body will be used for http authorization request (#27228)

jewertow:
This is a backport of an upstream bug fix to ext_authz, which does not pass a request body to an authorization server when `Content-Type: multipart/form-data` and `pack_as_bytes: true`.


twghu (approved):

twghu:
This was from upstream commit 0d19dcf255.

PR #307, opened 2024-02-21
Fix for OSSM-5962

twghu:
Added cipher check.


tedjpoole (approved):

twghu:
I've run tests locally outside of the bazel test env and the tests pass.

We can override ?

twghu:
One fix for timeouts I've tested is excluding the timeout sensitive tests from the main batch and running individually at the end.

dcillera:
Decided to override the tests to provide a fixed proxy and to investigate test TIMEOUT issue meanwhile.

PR #309, opened 2024-02-27
Attempt to remove time out error on tests

twghu:
Some bazel tests were consistently failing with TIME OUT errors.

Breaking up testing into blocks addresses this issue when building locally.

Breaking into blocks will also make test run/management easier.

### DIFF
--- a/maistra/run-ci.sh
+++ b/maistra/run-ci.sh
@@ -23,6 +23,7 @@ bazel-bin/source/exe/envoy-static --version
 # The following build step helps reduce resources usage
 # by compiling tests first.
 # Build tests
+
 time bazel build \
   ${COMMON_FLAGS} \
   --build_tests_only \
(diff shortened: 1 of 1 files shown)


maistra/run-ci.sh
```diff
@@ -23,53 +23,76 @@ bazel-bin/source/exe/envoy-static --version
...
-  --build_tests_only \
-  -- \
-  $test 
-done
+# --cache_test_results=no \
+
+# Run the tests in smaller blocks rather than attempt to 
+# share resources to run //test/... 
+# This is a remedy for the tests that would time out.
+run_tests "//test/common/..."
```
jwendell, line 49: Rather than hard coding their names, it would be better to have a for loop over the results of something like `find test -type d`.

twghu: Some folders are not testable individually and the point of the explicit list is to actually identify which works and which typically fails.

jwendell: I see. If it helps, we could have a "blacklist" of folders that would be ignored in such for loop, similar to what we have in this script: https://github.com/maistra/proxy/blob/89231677979b3d09186ce958bc68685e5e8129c7/maistra/scripts/update-deps.sh#L33

This could be done in a follow up though, as an improvement. The idea is that by not hard coding test/folder names we would be future-proof.

maistra/run-ci.sh
```diff
@@ -23,53 +23,76 @@ bazel-bin/source/exe/envoy-static --version
...
+run_tests "//test/extensions/common/..."
+run_tests "//test/extensions/compression/..."
+run_tests "//test/extensions/config/..."
+run_tests "//test/extensions/config_subscription/..."
+run_tests "//test/extensions/filters/common/..."
+run_tests "//test/extensions/filters/http/... -//test/extensions/filters/http/kill_request:crash_integration_test"
+run_tests "//test/extensions/filters/http/kill_request:crash_integration_test"
+run_tests "//test/extensions/filters/listener/... -//test/extensions/filters/listener/original_dst:original_dst_integration_test"
+# The following test only fails locally, passes on CI.
+#run_tests "//test/extensions/filters/listener/original_dst:original_dst_integration_test"
```
jwendell, line 66: enable it?

maistra/run-ci.sh
```diff
@@ -23,53 +23,76 @@ bazel-bin/source/exe/envoy-static --version
...
+run_tests "//test/extensions/udp_packet_writer/..."
+run_tests "//test/extensions/upstreams/..."
+run_tests "//test/extensions/watchdog/..."
+
+run_tests "//test/fuzz/..."
+run_tests "//test/integration/..."
+run_tests "//test/mocks/..."
+run_tests "//test/server/..."
+run_tests "//test/test_common/..."
+run_tests "//test/tools/..."
```
jwendell, line 97: do we need to run the `//test` folder as well?

maistra/run-ci.sh
tedjpoole: @dcillera we should run the `original_dst_integration_test` test right?

dcillera: I confirm

maistra/run-ci.sh
dcillera: Do you mean that there are further tests besides those listed in the sub-directories above?

jwendell: No. I asked if there are tests in the root `//test` folder that should run. Note that I did not write `//test/...`

dcillera: It seems no test targets are present

twghu:
Closed for https://github.com/maistra/envoy/pull/310

